PR #8, opened 2024-07-18
Icons statics: Don't use version in local webserver path

iLLiCiTiT:
## Description
Don't use addon version when supplying statics on localhost.

## Additional information
To be able to use the uri one has to know which addon version is used, which is not really necessary for localhost, which has to run in tray, which can use only one version of an addon at a time.


BigRoy:
I'm not sure how to best test this and what it may effect potentially.

64qam (approved):

PR #12, opened 2024-07-29
Add a Debug Shell launcher which allows to launch a terminal into environment of an application

BigRoy:
## Changelog Description

Implements a "Shell" launcher action:
image

When clicked allows you to pick the shell context to enter (For which application):

image

From that shell (when launched into Maya app for example) I can directly type `mayapy` to launch Maya within the environment of the Maya application:

image

## Additional info

Note: currently always runs `cmd` (Windows command prompt) which for now makes this Windows only when testing. We'll need to implement the following TODO to solve that:

- [x] Expose 'defaults' for the terminal to use in studio settings
    - This could be a **Terminal** application. Or are there reasons not to do that @iLLiCiTiT?
    - Other approaches to figuring out the best 'shell' to launch for a user is more than welcome. @dee-ynput  :)
- [x] Allow user-specific overrides for the terminal to use?
    - What type of settings would that need?
- [x] Use the first matching executable for current platform as defined by settings

## Testing notes:

1. Launch tray
2. Use the launcher action

---

AY-1261


client/ayon_applications/plugins/launcher_actions/debug_shell.py
```diff
@@ -0,0 +1,116 @@
...
+
+def get_application_qt_icon(
+        application: Application
+) -> Optional[QtGui.QIcon]:
+    """Return QtGui.QIcon for an Application"""
+    # TODO: Improve workflow to get the icon, remove 'color' hack
+    from ayon_core.tools.launcher.models.actions import get_action_icon
+    from ayon_core.tools.utils.lib import get_qt_icon
+    application.color = "white"
+    return get_qt_icon(get_action_icon(application))
```
iLLiCiTiT, line 24:
```suggestion
    icon = action.icon
    if not icon:
        return QtGui.QIcon()
    icon_filename = os.path.basename(icon)
    icon_filepath = os.path.join(APPLICATIONS_ADDON_ROOT, "icon", icon_filename)
    if os.path.exists(icon_filepath):
        return QtGui.QIcon(icon_filepath)
    return QtGui.QIcon()
```

client/ayon_applications/plugins/launcher_actions/debug_shell.py
```diff
@@ -0,0 +1,116 @@
...
+    icon = "terminal"
+    color = "#e8770e"
+    order = 10
+
+    def is_compatible(self, selection) -> bool:
+        return all((
+            selection.is_task_selected,
+            selection.is_folder_selected,
+            selection.is_project_selected,
+        ))
```
iLLiCiTiT, line 40: You can't select task without having selected folder or project.
```suggestion
        return selection.is_task_selected
```

iLLiCiTiT:
I can do code review, but don't understand the use-case so I can't review that part.

client/ayon_applications/plugins/launcher_actions/debug_shell.py
BigRoy: Doesn't this make it:
- fail to use the resource / icons cache?
- fail to allow awesome-font icons or alike (and its coloring) if anything _would/could_ define that?

dee-ynput:
> - Expose 'defaults' for the terminal to use in studio settings
>     - This could be a Terminal application. Or are there reasons not to do that @iLLiCiTiT?

This sounds appropriate. But wouldn't that create a loop? Like opening a shell in the environment of a shell 😃


>     - Other approaches to figuring out the best 'shell' to launch for a user is more than welcome. @dee-ynput :)
If we manage to have it defined as "Terminal" app I think we cover all our needs (platform executable, extra env, and cli args)

BigRoy:
> This sounds appropriate. But wouldn't that create a loop? Like opening a shell in the environment of a shell 😃

I'd say it'd then open a shell like:
- Get terminal env
- Overlay application env
- Launch the terminal with that env

> If we manage to have it defined as an identified variant of a "Terminal" app I think we cover all our needs (platform executable, extra env, and cli args).

Does this also automatically expose this as user-specific overrides somewhere? I can't recall seeing that. Say one particular dev wanted to use a different terminal app.

iLLiCiTiT:
> This could be a Terminal application. Or are there reasons not to do that @iLLiCiTiT?

Not sure what you mean by `Terminal` application, but each OS have way how to get default terminal. On windows can be used `wt.exe`. On macOs and linux it should be defined under env variable `SHELL` (this need validation).

NOTE: macOs requires `["open", "-na", os.getenv("SHELL"), subprocess.list2cmdline(args)]`.

m-u-r-p-h-y (approved):
works as expected 
image


I would also make this tool available only to admins so the Launcher UI is not cluttered with some unnecessary tools for artists. (or making it configurable in settings with default being off)

BigRoy:
1. [ ] We may need a Mac and Linux test here as well. If that works out of the box, then making the 'terminal' itself configurable for a specific user can be a follow up PR. @Lypsolon Linux? :) And who can test on Mac?
    - [x] Windows 
    - [ ] Linux
    - [ ] Mac
3. [ ] Allow to enable it only for admin or only in dev mode via settings (however, LauncherActions currently don't receive any project settings so querying that each time may be a bit slow?) We may also want to make that a separate "feature" and hence separate PR. @iLLiCiTiT thoughts? I'd also argue that it'll always appear near the end of the list due to the order `10` and the client who requested a feature like this may be a TD who's not necessarily "admin" in the project either?

m-u-r-p-h-y:
Or just an idea to treat it as a normal App, which can be enabled, disabled, configured, having variants for different shells (cmd, powershell, bash, tcsh etc.)

image

BigRoy:
> Or just an idea to treat it as a normal App, which can be enabled, disabled, configured, having variants for different shells (cmd, powershell, bash, tcsh etc.)
> 
> image

Yes - that's what @dee-ynput proposed, but that wouldn't technically allow it to be overridden _per user_, right?

MustafaJafar:
> If we manage to have it defined as an identified variant of a "Terminal" app I think we cover all our needs (platform executable, extra env, and cli args).

Should that setting exists in core settings beneath `Tools/Launcher Actions/Debug Shell`
-  in studio settings with project overrides ? (won't be per user)
- OR, in core site settings ? (per site id(machine), but no project overrides).
OR
Should it fetch the executable from a terminal app ? (which adds one more app in the launcher UI ,and not per user)

MustafaJafar:
> I would prefer it to be:
> 
>     * Studio provides defaults
> 
>     * User allows to override to pick another shell if they really want.

I like this idea but, it will make the debug action works in two step:
1. select your favorite shell/terminal
2. select your favorite app to debug

BigRoy:
> 2\. select your favorite app to debug

Not if that would be a user specific setting in the frontend, like a site-override or whatever. Didn't we have user settings?

MustafaJafar:
> Not if that would be a user specific setting in the frontend, like a site-override or whatever. Didn't we have user settings?

let me quote: 
No, we only have 4 scopes of settings:
- studio (studio settings) : achieved by specifying it in scope list.
- project (project settings): achieved by specifying it in scope list.
- site (project site settings): achieved by specifying it in scope list.
- (studio site settings): Currently achieved by adding settings to site_settings_model attribute  in the addon class

BigRoy:
I've now added dedicated "shell" application settings and use that to launch the terminal/console/shell with the application's environment. (It should actually, I believe - _also_ apply the application env as defined in `applications/shell`)

Note that it's hardcoded for now to use the `shell/main` entry as I wasn't sure how to best allow an artist to pick the particular version they want to launch of the shell terminal. Maybe what @MustafaJafar explained would be best - that if there are multiple variants configured that it'd just list them all and allow you to choose - otherwise if there's just one, use that. If there's none... then it'd error?

m-u-r-p-h-y:
> Yes - that's what @dee-ynput proposed, but that wouldn't technically allow it to be overridden _per user_, right? Say, one dev prefers a different terminal.

the same story as with any other artist. One prefers Blender the other Maya for modelling. What is the difference? 
image

MustafaJafar:
> > Yes - that's what @dee-ynput proposed, but that wouldn't technically allow it to be overridden _per user_, right? Say, one dev prefers a different terminal.
> 
> the same story as with any other artist. One prefers Blender the other Maya for modelling. What is the difference?

powershell can be tricky, because, I had to download a new version of it. and I'm not sure if users will keep the default installation path.

BigRoy:
> > Yes - that's what @dee-ynput proposed, but that wouldn't technically allow it to be overridden _per user_, right? Say, one dev prefers a different terminal.
> 
> the same story as with any other artist. One prefers Blender the other Maya for modelling. What is the difference? image

Thanks @m-u-r-p-h-y this should now work - and user should be able to pick the relevant configured shell application if there's more than one.

BigRoy:
For whatever reason the Application icon doesn't work for me in the launcher. Does it need to be in `public/` or in `client/ayon_applications/icons`, or both? @iLLiCiTiT

MustafaJafar (commented):
Funny thing, When using the Shell main, doesn't open in a new tab. 
Maybe we need to add `start` when calling `cmd.exe`  as mentioned [here](https://stackoverflow.com/questions/303838/create-a-new-cmd-exe-window-from-within-another-cmd-exe-prompt)
image

MustafaJafar (commented):
when clicking the debug shell action, ayon complaints that it can't find the icon of shell main. (tested  running launcher in production and dev) same happens. it was already mentioned here https://github.com/ynput/ayon-applications/pull/12#discussion_r1696725615

image

MustafaJafar (approved):
This is cool. It works. I love it.
It doesn't only set environment variables defined in the applications addon but also it triggers the Hooks.
image

client/ayon_applications/plugins/launcher_actions/debug_shell.py
MustafaJafar: Hey,
I debugged this for a while, I've an alternative solution.

```suggestion
    from ayon_core.tools.launcher.models.actions import get_action_icon
    from ayon_core.tools.utils.lib import get_qt_icon
    application.color = "white"

    icon = get_qt_icon(get_action_icon(application))
    sizes = icon.availableSizes()
    if sizes:
        if not (sizes[0].toTuple() > (1, 1) ):
            icon_filename = os.path.basename(application.icon)
            icon_filepath = os.path.join(
                APPLICATIONS_ADDON_ROOT, "icons", icon_filename)
            if os.path.exists(icon_filepath):
                return QtGui.QIcon(icon_filepath)
            return QtGui.QIcon()
    return icon
```
image

MustafaJafar: or maybe 
```suggestion
    from ayon_core.tools.launcher.models.actions import get_action_icon
    from ayon_core.tools.utils.lib import get_qt_icon
    application.color = "white"
    icon = get_qt_icon(get_action_icon(application))
    sizes = icon.availableSizes()
    if not sizes:
        return QtGui.QIcon()
        
    if sizes[0].toTuple() > (1, 1):
        return icon
        
    icon_filename = os.path.basename(application.icon)
    icon_filepath = os.path.join(
        APPLICATIONS_ADDON_ROOT, "icons", icon_filename)
    if os.path.exists(icon_filepath):
        return QtGui.QIcon(icon_filepath)
    
    return QtGui.QIcon()
```

BigRoy: Had a look myself - it basically failed to resolve the filepath to the icon correctly because it looked in `ayon_core` resources for the icon, but it's now in ayon applications and I think that's why @iLLiCiTiT provided me the suggestion.

I've now pushed a fix.

BigRoy:
> Funny thing, When using the Shell main, doesn't open in a new tab. Maybe we need to add `start` when calling `cmd.exe` as mentioned [here](https://stackoverflow.com/questions/303838/create-a-new-cmd-exe-window-from-within-another-cmd-exe-prompt) But, we can leave it for another PR. image

Yes correct - the actual shell application launching doesn't work because it doesn't open a new console.

I had to force these creationflags to make it work: https://github.com/ynput/ayon-applications/blob/2d58dcd1611ba35ad089f28cd40717f9f8f4bd73/client/ayon_applications/plugins/launcher_actions/debug_shell.py#L176 for what I'm using to run the shell.

server/applications.json
```diff
@@ -1302,6 +1302,31 @@
                 }
             ]
         },
+        "shell": {
```
iLLiCiTiT, line 1305: Why there is setting for this?

BigRoy: Thanks. @m-u-r-p-h-y @dee-ynput your thoughts?

I still feel, based on the comments above by everyone, we will need a way to allow the user to pick what shell to trigger that works for their needs.

Also, the assumption that this PR may solve the client's needs based on their ticket @m-u-r-p-h-y will still need to be confirmed. That may also clarify any customization needs as well. So may be good to follow up on that before we continue here?

---

Side note: I do know that @Lypsolon also requested this feature to be able to quickly access a terminal in the environment of the host - so it would maybe still solve at least a request from him.

iLLiCiTiT: We should use SHELL on linux and mac. On windows`wt.exe` if exists otherwise `cmd.exe` and you have to know which is used so you can e.g. add  `/K` for cmd.

BigRoy: Side note: For me `wt.exe` actually didn't work even though I do have the executable. I was unable to run it with an initialized environment it seemed - it'd always use the system environment and also ignore `cwd`.

MustafaJafar: > I still feel, based on the comments above by everyone, we will need a way to allow the user to pick what shell to trigger that works for their needs.

let share my user story, I hope it helps: 

Personally, I'd expect this to exist in studio site settings or application settings.
`studio site settings` allows me to have my own customization (which is none atm) at the cost of doing this for every machine I use to log into the system. as I user I can have multiple site ids.
`application settings` is pretty straight forward. it also allows me to open a console in some host specific. which I may need for testing purposes. 
e.g. in shell application(currently launched via the debug shell), I'm able to access ayon libs!  e.g. `ayon_core`
image


Regarding the customization per users, on my side, I'm just a regular guy, who have a basic setup to get the job done.
I'd either go with `cmd` or `pwsh` (we don't need to specify the whole path in the shell app settings because I already added it to PATH on my machine)
Also, I don't have any customization therefore, shell application is fine.

dee-ynput: > Why there is setting for this? Please don't, it's not used as application, you want to run other application with shell.

Yes it is an application.
It just happens to be an application which **can** run other applications.
Can you elaborate on why we shoud not do that @iLLiCiTiT ?

I have a feeling there's a miss understanding of the purpose. Maybe because of different meaning we have for the words `shell`, `terminal`, `console` ?

The goal is to open a console where the user can do what they want.
(Maybe you're thinking about launching eg Maya and have a terminal showing up to see log and such?)

> We should use SHELL on linux and mac.

Yes, `$SHELL` is probably the default value for linux and mac.
But the studio and the projects must be able to configure something else, and variants too.

Guys, let's ship this, get some feedback, and then come back to the discussion.

BigRoy: @MustafaJafar I've fixed the regular app launch with this https://github.com/ynput/ayon-applications/pull/12/commits/144163d05a2a714f11e807fa0d29d3d7feca88c8 comit.

That commit should also fix launch on Linux and Max because apparently `subprocess.CREATE_NEW_CONSOLE` does not exist on Linux. It's windows only?

@Lypsolon would be great if you could just give this a Linux test-drive.
And I'm clueless as to who would be capable of testing Mac.

MustafaJafar (approved):
This is cool! 
I can launch the shell app regularly. 
image

I don't think there's much value in launching the shell app from the debug action. 
It only sets the directory to the my task path.

---
Launching apps. 
I really like it. and like how the startup scripts kicks in when calling some terminal apps.
~~I'm here struggling with running `usdview` shipped with houdini's USD build~~ but I like how `$JOB` was updated! 
I also ran houdini and still can't tell a difference from running it directly via launcher.

image

Lypsolon (commented):
A list of things I found 

- the shell app shows me the option to launch houdini19.5, but I did not configure it 
Screenshot from 2024-08-05 15-13-03

- after adding Houdini 19.5 to the executable, the shell still won't show. 


with the current PR i was not able to get a shell working. 

```
Only one shell application variant is configured. Defaulting to Shell Main
Retrieving environment for: Houdini 19.5..
>>> [  Colorspace management is disabled globally.  ] 
>>> [  Current context does not have any workfile yet.  ] 
>>> [  Initializing USD asset resolver for [ houdini/19-5 ] .  ] 
>>> [  Asset resolver houdini/19-5 initiated.  ] 
Appending to PATH: /opt/hfs19.5.974/bin
Setting Work Directory: /home/lyonh/Documents/AyonServerDevMain/Usd_Base/versiond_asset_test/work/Maya
Launching shell in environment of Houdini 19.5..
>>> [  Launching "shell/main" with args (1): ['/usr/bin/zsh']  ] 
```
ps: the path `/usr/bin/zsh` is correct and works 


Testing Notes: 
- cloned your repo 
- switched branch 
- python create_package.py 
- installed in the server. 
- setup Houdini again (after the first test, images atop explain more)
- pressed the buttons in the launcher

BigRoy:
Thanks @Lypsolon 

Could you confirm a few things?

Houdini 19.5 is configured and enabled in the Project's anatomy. That is why it is listed. Unlike the default applications behavior it does not hide applications you do not have installed but are enabled/configured in your project. I am pretty sure that is your case.

We can add functionality that makes it behave similar to other apps so that it only lists any applications that exist if that settings is enabled in Application settings. (It's at the far bottom of the application studio settings I believe)

Then two, the shell not showing may be equivalent to why it failed initially on Windows, on Windows opening a shell via subprocess required us to disconnect stdin, stdout and launch with subprocess `CREATE_NEW_CONSOLE` flag which is not available under Linux. So what would be interesting to test is whether `subprocess.Popen("/usr/bin/zsh")` works for you or whether you also need to pass additional arguments for it to correctly open a new console for the command prompt. What works for you?

Lypsolon:
> Houdini 19.5 is configured and enabled in the Project's anatomy. That is why it is listed. Unlike the default applications behavior it does not hide applications you do not have installed but are enabled/configured in your project. I am pretty sure that is your case.

it looks like it, tbh. 

> We can add functionality that makes it behave similar to other apps so that it only lists any applications that exist if that settings is enabled in Application settings. (It's at the far bottom of the application studio settings I believe)

I think that might be good to do as this behaviour could be confusing. at the same time I think we should put it on the bottom of the list. 

> Then two, the shell not showing may be equivalent to why it failed initially on Windows, on Windows opening a shell via subprocess required us to disconnect stdin, stdout and launch with subprocess `CREATE_NEW_CONSOLE` flag which is not available under Linux. So what would be interesting to test is whether `subprocess.Popen("/usr/bin/zsh")` works for you or whether you also need to pass additional arguments for it to correctly open a new console for the command prompt. What works for you?

ähh no, it doesn't because the command will open a shell, not a terminal window. 
So, at the end of the day, yes, it will open a new shell inside the current shell, but no window will pop up. 
To open a terminal window, call the associated system from the desktop/window manager. 
in gnome that would be `subprocess.run(["gnome-terminal"])` 
PS: there might be other ways to do the same, but I only know this one.

MustafaJafar:
@Lypsolon 
does the debug shell launcher work as expected but it doesn't open a new window ? 
or 
the shell app is the one that doesn't open in a new window ?

Lypsolon:
> > @Lypsolon does the debug shell launcher work as expected but it doesn't open a new window ? or the shell app is the one that doesn't open in a new window ?
> > I'm just trying to figure out what is still missing ?
> 
> I suppose both won't work since the issue is the same, we need a way to spawn of a new console for the shell. We just need to find the canonical way to open a new terminal window on Linux similar to the command on Windows, hopefully one that doesn't have to be gnome specific. Or we just tell Linux admins to set up the executable paths to include the gnome-terminal command.
> 
> Also assume that Lyon never added the Shell app to the project so didn't use just the shell (without any app environment other than its own).

I tested adding it to the project, but to no surprise, I got the same result.


```
# A list of terminal emulators that might be installed.
# This allows admins to add extras, and we could cover a few defaults.
# This also allows the addition of more defaults as requests come in. 
terminal_emulators_addon_settings = [
    "gnome-terminal",
    "x-terminal-emulator",
    "konsole",
    "xfce4-terminal",
    "lxterminal",
    "mate-terminal",
    "tilix",
    "deepin-terminal",
    "terminator",
    "alacritty",
    "st",
    "urxvt",
    "xterm"
]

def get_terminal():
    for terminal in terminal_emulators:
        if shutil.which(terminal):
            return terminal
    return None

subprocess.run([get_terminal()])

```

iLLiCiTiT:
I'm not sure if it I mentioned it already, but I'm really against having Shell in settings of applications, it is just not something that should be modified on studio level.

server/applications.json
iLLiCiTiT: It is not application like maya, nuke or something like that. We've already removed shell from setting once because it raised exactly the issues we're mentioning here. The terminal cannot be set "universaly" as it is machine specific. Also it doesn't make sense to have variants as we have it.

We should have auto-detect logic, and maybe site settings for shell, but it is 100% not the same application as we have the others.

We need shell for running specific action, that will start terminal window which has set environment as other application, not run "shell as application".

MustafaJafar:
A fun fact, I thought this PR was merged already :D 
Anyways, should we merge it and fix the issue later ? 

The issue :  quoted from https://github.com/ynput/ayon-applications/pull/12#issuecomment-2271166843
```
find the canonical way to open a new terminal window on Linux similar to the command on Windows,
```

Lypsolon:
> A fun fact, I thought this PR was merged already :D Anyways, should we merge it and fix the issue later ?
> 
> The issue : quoted from [#12 (comment)](https://github.com/ynput/ayon-applications/pull/12#issuecomment-2271166843)
> 
> ```
> find the canonical way to open a new terminal window on Linux similar to the command on Windows,
> hopefully one that doesn't have to be gnome specific. 
> Or we just tell Linux admins to set up the executable paths to include the gnome-terminal command.
> ```

correct me if I'm wrong, but currently, we do not support Linux at all, so if we want to merge it, I believe we should implement some solution or mark the feature as Windows only.

MustafaJafar:
> correct me if I'm wrong, but currently, we do not support Linux at all, so if we want to merge it, I believe we should implement some solution or mark the feature as Windows only.

As far as I know, this feature works pretty well on windows. 
about the Linux support, I'm not sure about its status, so I'll leave it for @iLLiCiTiT 
Also, tagging @m-u-r-p-h-y could you tell us if marking this as `windows only` satisfies the client request ?

iLLiCiTiT:
There is conversation in the code. I would like to remove terminal from settings.

server/applications.json
BigRoy: @dee-ynput @antirotor Please provide your opinions so we can **force** this into **removal of terminal/shell** from application settings OR autodetect it in a different way.

I suspect there's _still_ merit to just a 'shallow' shell within the AYON environment without the context of other applications which may warrant exposing a "shell" application but I'm not convinced either. Please take into account @iLLiCiTiT 's points above ☝️ and let's get this ball rolling.

iLLiCiTiT: > I think that having shell as "standard" application in AYON might be useful,

I really don't see any usecase when the terminal application would be used as "application" and would be useful. I would say that the action added in this PR is useful for developer, but terminal application is not.

If it would be for development, you can easily add it in additional applications. But I would not put it to settings to make it visible in applications enumerator.

> We need default/studio/project definition of what the shells are so we have some sane starting point. This should be defined as standard application in the application addon with variants and everything as was already shown in this discussion - it allows running the shell console in selected context.

I've mentioned it already but I'm strongly against having it in `Applications` settings.

Even this PR is not using the shell application as application, but is adding another action, different from application launch action to launch the terminal.

Ad terminal choice, on macOs it is `Terminal.app` (probably no other choice?). On windows it can be powershell 7, powershell 6, cmd, git console, on linux it can be many many applications. On linux you probably should know which exactly terminal it is so you know how to work with it and start the session correctly. Meaning it is not really studio specific but machine specific. By default we should do as much as we can to start correct console, but only if user don't like it, he should set it in site settings, but not in studio settings.

Right now I'm probably only one who thinks this is bad idea, so I'll probably loose this fight, just wanted to mention this (sorry).

dee-ynput:
My 2cents:

### About having a "Shell" in application
It's totally relevant. And for those who don't use it, it's like having blender there even though you don't use it. I do see how problematic it is, but that's true for every entry so let's not make this one any special. We need it in order to have a official shell app which is used by other things (like the open this app in a shell action, but more will come).

Yes, it's a developer tool. But AYON is made for developers too.

**Let's add it.**

### About configuring a shell per user/site
We don't need that. There would be one or two, probably never more than 3 supported in a studio. Let's make them variant of the application.
If a project wants to add more, additionnal app would help. But we'll deal with this request IF/WHEN it comes from users.

### About using the default shell vs an App defined it settings:
> I don't see the usecase where would TD set "studio" terminal and it would be useful.

Here is the use case: When you're a TD debugin someone at his station you need to access a standardized shell to start digging around. Artist won't setup their shell env, and even for the ones doing it you don't want to deal with their custom config: you need an official one.

### About using wt/powershell/$SHELL/Terminal.app/etc...
Those are good defaults. But all platforms / distribution have alt consoles that a studio may want to use. Settings should give freedom to configure that, even if 90% of the time it will stay on default.

iLLiCiTiT:
Ok, so I lost my battle, now let's do the real review...

client/ayon_applications/plugins/launcher_actions/debug_shell.py
```diff
@@ -0,0 +1,175 @@
...
+    if os.path.exists(icon_filepath):
+        return get_qt_icon({"type": "path", "path": icon_filepath})
+    return QtGui.QIcon()
+
+
+class DebugShell(LauncherAction):
+    """Run any host environment in command line."""
+    name = "debugshell"
+    label = "Shell"
+    icon = "terminal"
```
iLLiCiTiT, line 39: Is this font-awesome icon? If yes, add `fa5.` prefix please.

client/ayon_applications/plugins/launcher_actions/debug_shell.py
```diff
@@ -0,0 +1,175 @@
...
+from ayon_core.style import load_stylesheet
+from ayon_core.tools.utils.lib import get_qt_icon
+
+
+def get_application_qt_icon(application: Application) -> Optional[QtGui.QIcon]:
+    """Return QtGui.QIcon for an Application
+
+    Note: This forces the icons to be search in `ayon_applications/icons`
+        folder on client side and mimics what Ayon applications addon's
+        `get_app_icon_path` method does.
```
iLLiCiTiT, line 22: If that's the case then please use `get_app_icon_path` from addon, or move the function to `utils.py`, because right now this is different logic from what is in addon method.

client/ayon_applications/plugins/launcher_actions/debug_shell.py
```diff
@@ -0,0 +1,175 @@
...
+    def process(self, selection, **kwargs):
+        # Get cursor position directly so the menu shows closer to where user
+        # clicked because the get applications logic might take a brief moment
+        pos = QtGui.QCursor.pos()
+        application_manager = ApplicationManager()
+
+        # Choose shell
+        shell_applications = self.get_shell_applications(application_manager)
+        if len(shell_applications) == 0:
+            raise RuntimeError(
```
iLLiCiTiT, line 55: `RuntimeError` is kinda cruel.
```suggestion
            raise ValueError(
```

client/ayon_applications/plugins/launcher_actions/debug_shell.py
```diff
@@ -0,0 +1,175 @@
...
+            return result.data()
+
+    @staticmethod
+    def get_project_applications(application_manager: ApplicationManager,
+                                 project_entity: dict) -> list[Application]:
+        """Return the enabled applications for the project"""
+        # Filter to apps valid for this current project, with logic from:
+        # `ayon_core.tools.launcher.models.actions.ApplicationAction.is_compatible`  # noqa
+        applications = []
+        for app_name in project_entity["attrib"].get("applications", []):
```
iLLiCiTiT, line 140: Function `get_applications_for_context` should be used.

iLLiCiTiT:
Can we rename all the `shell` to `terminal`? All platforms are calling it terminal (even windows).

server/applications.json
```diff
@@ -1302,6 +1302,31 @@
...
+            "enabled": true,
+            "label": "Shell",
+            "icon": "{}/app_icons/shell.png",
+            "host_name": "shell",
```
iLLiCiTiT, line 1309: It does not has host integration, should not have host name.

server/applications.json
```diff
@@ -1302,6 +1302,31 @@
...
+            "environment": "{}",
+            "variants": [
+                {
+                    "name": "main",
+                    "label": "Main",
+                    "use_python_2": false,
+                    "executables": {
+                        "windows": ["cmd.exe"],
+                        "darwin": ["open"],
```
iLLiCiTiT, line 1318:
```suggestion
                        "darwin": ["Terminal.app"],
```

server/applications.json
```diff
@@ -1302,6 +1302,31 @@
...
+                        "linux": ["{SHELL}"]
+                    },
+                    "arguments": {
+                        "windows": [],
+                        "darwin": ["-na", "{SHELL}"],
```
iLLiCiTiT, line 1323:
```suggestion
                        "darwin": [],
```

client/ayon_applications/plugins/launcher_actions/debug_shell.py
```diff
@@ -0,0 +1,175 @@
...
+        return applications
+
+    @staticmethod
+    def get_shell_applications(application_manager) -> list[Application]:
+        """Return all configured shell applications"""
+        # TODO: Maybe filter out shell applications not configured for your
+        #  current platform
+        return list(application_manager.app_groups["shell"].variants.values())
+
+    def launch_app_as_shell(
```
iLLiCiTiT, line 155: Maybe better?
```suggestion
    def launch_terminal_with_app_context(
```

server/applications.json
```diff
@@ -1302,6 +1302,31 @@
                 }
             ]
         },
+        "shell": {
+            "enabled": true,
+            "label": "Shell",
+            "icon": "{}/app_icons/shell.png",
```
iLLiCiTiT, line 1308:
```suggestion
            "icon": "shell.png",
```

iLLiCiTiT: My bad, this should not be in settings at all.

server/applications.json
```diff
@@ -1302,6 +1302,31 @@
                 }
             ]
         },
+        "shell": {
+            "enabled": true,
+            "label": "Shell",
```
iLLiCiTiT, line 1307: `"label"` nad `"icon"` should be moved to constants.

client/ayon_applications/plugins/launcher_actions/debug_terminal.py
```diff
@@ -0,0 +1,178 @@
...
+    ApplicationManager,
+    APPLICATIONS_ADDON_ROOT
+)
+from ayon_applications.utils import get_app_environments_for_context
+from ayon_core.pipeline import LauncherAction
+from ayon_core.style import load_stylesheet
+from ayon_core.tools.utils.lib import get_qt_icon
+
+
+def get_application_qt_icon(application: Application) -> Optional[QtGui.QIcon]:
```
iLLiCiTiT, line 17: Just reminder, there already is function to get icon path, and there should be only one. I would move it from addon method to `utils.py` so it can be used anywhere.

BigRoy: Lovely - where is the 'already existing one' that I should be adopting?

client/ayon_applications/plugins/launcher_actions/debug_shell.py
BigRoy: Thanks, done with https://github.com/ynput/ayon-applications/pull/12/commits/6d139e86ca2bce0a7939e6a4f7f0521466ca65d1

client/ayon_applications/plugins/launcher_actions/debug_terminal.py
BigRoy: Did you mean `get_app_icon_path` that you mentioned [here](https://github.com/ynput/ayon-applications/pull/12#discussion_r1738852629)?

This seems to be [here](https://github.com/ynput/ayon-applications/blob/144163d05a2a714f11e807fa0d29d3d7feca88c8/client/ayon_applications/addon.py#L122-L138) on the Applications addon object. From the Launcher action, what is the best efficient way to access the addon instance? Do I need to initialize a full `AddonsManager` instance?

client/ayon_applications/plugins/launcher_actions/debug_shell.py
BigRoy: Fixed with https://github.com/ynput/ayon-applications/pull/12/commits/2432635862b192af2ef3f2781b025982d8fdf4e7

client/ayon_applications/plugins/launcher_actions/debug_terminal.py
iLLiCiTiT: Yes, that's why I would move the logic to `utils.py` and import/use it where is needed.

BigRoy:
@Lypsolon I've tried to change the defaults for the terminal on Linux, to instead use the list you had provided. Likely still won't work, but if there's any chance you could try with new `ayon_applications` build package and trying to run this... that'd be lovely! ❤️ 

Also, @iLLiCiTiT will bitch to me about [this insane list](https://github.com/ynput/ayon-applications/pull/12/commits/c2b6eef10d03c008cc8965fb7b258d3f2ae6ae71). And @dee-ynput I suppose too. So feel free to fight about it - just wanted to throw this in there for @Lypsolon to double check whether this would actually be a suitable workflow for Linux, potentially. (Which I suppose @dee-ynput can test nowadays too!?)

client/ayon_applications/plugins/launcher_actions/debug_terminal.py
BigRoy: Implemented with https://github.com/ynput/ayon-applications/pull/12/commits/d41e43ee1c715ed3da074c5afeb9440b3df8f1fc **but** the `get_application_qt_icon` logic is still more involved and as you said may warrant moving to that `utils`. Moving that to `utils` may mean that the Addon itself also imports `qtpy` - which is likely not what we want?

Also, I just want to emphasize that there's **confusingly** also a `get_app_icon_path` function in `ayon_core.style`, see: https://github.com/ynput/ayon-core/blob/0701238b531f052bb526429186c263ac31fa1c38/client/ayon_core/style/__init__.py#L208-L210

I guess we'll still need to refactor/cleanup some more. Any pointers on how you'd like me to proceed?

iLLiCiTiT: I didn't mean `get_application_qt_icon`, I meant just the 

> Also, I just want to emphasize that there's confusingly also a get_app_icon_path function in ayon_core.style, see: https://github.com/ynput/ayon-core/blob/0701238b531f052bb526429186c263ac31fa1c38/client/ayon_core/style/__init__.py#L208-L210

Should be removed in future, it is there for backwards compatibility.

client/ayon_applications/icons/terminal.png
iLLiCiTiT, line 1: This should be also added to `public` folder.

server/settings.py
```diff
@@ -42,6 +42,7 @@
     "zbrush",
     "equalizer",
     "motionbuilder",
+    "terminal"
```
iLLiCiTiT, line 45:
```suggestion
    "terminal",
```

iLLiCiTiT:
Missing icon and label in `client/ayon_applications/contants.py`.

iLLiCiTiT:
> Also, @iLLiCiTiT will bitch to me about [this insane list](https://github.com/ynput/ayon-applications/pull/12/commits/c2b6eef10d03c008cc8965fb7b258d3f2ae6ae71).

This is what I expected in the automated logic, so fine for me.

BigRoy:
> Missing icon and label in `client/ayon_applications/contants.py`.

Sorry, where? Not [here](https://github.com/ynput/ayon-applications/blob/7c9061b3d15e4a2ec1a8afb5345cdc68f7d90f48/client/ayon_applications/constants.py#L2-L1)?

iLLiCiTiT:
> Sorry, where? Not [here](https://github.com/ynput/ayon-applications/blob/7c9061b3d15e4a2ec1a8afb5345cdc68f7d90f48/client/ayon_applications/constants.py#L2-L1)?

Sorry I'm on different branch, in develop it is in `client/ayon_applications/defs.py`.

BigRoy:
> Sorry I'm on different branch, in develop it is in `client/ayon_applications/defs.py`.

Thanks, implemented with https://github.com/ynput/ayon-applications/pull/12/commits/00fe41ca9cb4eb0401fb8a7064f6663025b7ff04

server/settings.py
```diff
@@ -319,6 +320,10 @@ class ApplicationsSettings(BaseSettingsModel):
         default_factory=AppGroup, title="3DEqualizer")
     motionbuilder: AppGroup = SettingsField(
         default_factory=AppGroup, title="Motion Builder")
+    terminal: AppGroup = SettingsField(
+        default_factory=AppGroup,
+        title="Terminal",
+        description="Terminal shell application")
```
iLLiCiTiT, line 326:
```suggestion
        description="Terminal application")
```

iLLiCiTiT (approved):
If we really want to do this, then code is ok for me.

BigRoy:
> If we really want to do this, then code is ok for me.

🙈 🙉 🙊  - Thanks! Time will tell - and potentially hear us crying at night if we were wrong about this. But thanks again for pointing out your standpoint on this with the reviews, appreciate it!

---

@Lypsolon @dee-ynput ball is in your court to test the Linux side of things.

Lypsolon:
> > If we really want to do this, then code is ok for me.
> 
> 🙈 🙉 🙊 - Thanks! Time will tell - and potentially hear us crying at night if we were wrong about this. But thanks again for pointing out your standpoint on this with the reviews, appreciate it!
> 
> @Lypsolon @dee-ynput ball is in your court to test the Linux side of things.

i tried to test package creation and that works but after uploading the server restarts and dose not show the addon.

BigRoy:
> i tried to test package creation and that works but after uploading the server restarts and dose not show the addon.

Thanks. Checking..

Lypsolon (approved):
sorry it worked i just happend to merge ayon core into ayon applications while adding your repo as a remote my bad. 

i did a simple test and for me it works, i shall note this is only on Almalinux 9 but i approve the changes.

PR #15, opened 2024-08-02
Add Houdini default executable path examples for Mac and Linux

BigRoy:
## Changelog Description

Set up draft as defaults for Houdini executables on mac and linux

## Additional info

These paths may need to be confirmed to be correct by Linux/Mac users.

Based on e.g. https://www.sidefx.com/faq/question/install-linux/

## Testing notes:

1. Confirm the defaults
2. The new defaults should apply
3. Settings should work when loading new applications package


server/applications.json
```diff
@@ -656,8 +656,10 @@
                         "windows": [
                             "C:\\Program Files\\Side Effects Software\\Houdini 19.5.805\\bin\\houdini.exe"
                         ],
-                        "darwin": [],
-                        "linux": []
+                        "darwin": [
+                            "/Applications/Houdini/Houdini19.5.805/Frameworks/Houdini.framework/Versions/Current/Resources/bin/houdini"
+                        ],
+                        "linux": ["/opt/houdini19.5.805/bin/houdini"]
```
BigRoy, line 662: @Lypsolon any chance you can confirm this is e.g. matching what would be default path for houdini in your linux machine?

server/applications.json
```diff
@@ -656,8 +656,10 @@
                         "windows": [
                             "C:\\Program Files\\Side Effects Software\\Houdini 19.5.805\\bin\\houdini.exe"
                         ],
-                        "darwin": [],
-                        "linux": []
+                        "darwin": [
+                            "/Applications/Houdini/Houdini19.5.805/Frameworks/Houdini.framework/Versions/Current/Resources/bin/houdini"
```
BigRoy, line 660: There are also examples online that seem to refer to:
```
/Applications/Houdini/Houdini20.0.506/Frameworks/Houdini.framework/Resources/bin/houdini
```

I'm not sure what the good default is for Mac - so this answer may need to come from an active mac user :)

Lypsolon (commented):
The Linux defaults are reasonable for Houdini. but you should not point to a specific Houdini patch version. there is usually a symlink called hfs20.0 or hfs20.5, etc; I would say it is more likely that the symlink is set (via the Houdini launcher) than that the exact Houdini version is installed. 

ps: If a user changes the default install location, they can change the settings, I say. 

Also not sure if it should be part of this pr, but Houdini 20 and maybe even 20.5 would be nice

BigRoy:
> The Linux defaults are reasonable for Houdini. but you should not point to a specific Houdini patch version. there is usually a symlink called hfs20.0 or hfs20.5, etc; I would say it is more likely that the symlink is set (via the Houdini launcher) than that the exact Houdini version is installed.
> 
> ps: If a user changes the default install location, they can change the settings, I say.
> 
> Also not sure if it should be part of this pr, but Houdini 20 and maybe even 20.5 would be nice

Could you provide me the full default executable paths (and preferably even "suggest" them so we can finalize this PR?)

Lypsolon:
> executable

`/opt/hfs20.0` this is an path example. 
also there is an almalinux 9 machine in the ynput office if you want to test on Linux there is infrastructure for that. (its limited and i hope we get to a way better point but there is a baseline)

Lypsolon (commented):
just a few cosmetics. 
the Houdini versions are out of order 
19.5 19 18.5 20.5
thats a bit meh. can we have it 18.5 19 19.5 20 20.5 ? 

aaaand you are missing Hou20.0 no ?

BigRoy:
> thats a bit meh. can we have it 18.5 19 19.5 20 20.5 ?
> 
> aaaand you are missing Hou20.0 no ?

Yes, and yes. But this PR just adds paths for Linux and Mac - I did not touch the order.
I can shuffle it around and add Hou 20.0 of course - not sure why it was lacking upstream.

Lypsolon:
> > thats a bit meh. can we have it 18.5 19 19.5 20 20.5 ?
> > aaaand you are missing Hou20.0 no ?
> 
> Yes, and yes. But this PR just adds paths for Linux and Mac - I did not touch the order. I can shuffle it around and add Hou 20.0 of course - not sure why it was lacking upstream.

understandable. i think the shuffle and adding 20.0 would be great. 
if we dont want it in this PR thats fine. 

rest looks good.

BigRoy:
@Lypsolon can you approve this with a review? 

I can look into shuffling the order and add houdini 20.0 in a separate PR.

MustafaJafar:
Hey,

I followed along the discussion and checked the installation guides as well as some YT videos. 
So, there are installation directory and symlink to that directory (*which will be most likely overridden*) . 
e.g.
- Install directory: `/opt/hfs20.5.332/`
- Symlink:  `/opt/hfs/20.5/`

BigRoy:
> e.g.
> 
>     * Install directory: `/opt/hfs20.5.332/`
> 
>     * Symlink:  `/opt/hfs/20.5/`
> 
> 
> I wonder what do people in various studios mostly use?

Just for context - I actually changed this to the symlink one based on comments from @Lypsolon but I had the other one originally. Input of course is greatly appreciated!

MustafaJafar:
> other than that i believe that the symlinc is the best as its updates when you update via launcher etc.

yeah, but, if they are using Major.minor.patch version, it may fail, correct?

Lypsolon:
> > other than that i believe that the symlinc is the best as its updates when you update via launcher etc.
> 
> yeah, but, if they are using Major.minor.patch version, it may fail, correct?

what are you talking about ? 

when you install a version of Houdini on linux a symlinc fouler is created to point to the newest (i belive prod) version of Houdini you have installed. if you now update hou 20.5.01 to hou 20.5.02 the symlinc will be updated by the Houdini launcher. 

thats the idea behind the system that they implemented. 
it is posible that this is a houdini launcher installer funciton as i did not install houdini by hand for a long while. 

also its a lot more possible that some one installed a version of Houdini that we did not have in the defaults than that they dont have the symlinc

BigRoy:
> when you install a version of Houdini on linux a symlinc fouler is created to point to the newest (i belive prod) version of Houdini you have installed. if you now update hou 20.5.01 to hou 20.5.02 the symlinc will be updated by the Houdini launcher.
> 
> thats the idea behind the system that they implemented. it is posible that this is a houdini launcher installer funciton as i did not install houdini by hand for a long while.
> 
> also its a lot more possible that some one installed a version of Houdini that we did not have in the defaults than that they dont have the symlinc

Correct - but from a studio POV that may be less preferred. For context, we have multiple Houdini 19.5 and 20.0 versions in different productions. If using the symlink path, I'd then need to remap them manually. 

Also, I dislike it a tiny bit because it mismatches what Houdini and Mac does.

Lypsolon:
> Correct - but from a studio POV that may be less preferred. For context, we have multiple Houdini 19.5 and 20.0 versions in different productions. If using the symlink path, I'd then need to remap them manually.

if you use different versions with prjA and prjB you will need to remap / change it in the settings in all cases no meter what we put there. 
i believe question should be what is the most likely to be the default, if you run "none default" or multiple versions we will need to touch the settings in every case. 

> Also, I dislike it a tiny bit because it mismatches what Houdini and Mac does.

Houdini and Mac ? you mean windows and mac right ? 
with that one i get the point but would say a default should be as close to the default on the given system, i mean a mac user dosnt care what windows dose because why would he and vise versa. 
i believe the setting for linux should feel natural for the linux user, the same for windows and mac.

BigRoy:
> if you use different versions with prjA and prjB you will need to remap / change it in the settings in all cases no meter what we put there.

No you don't ;) you'd have a default template showing you how to set it to a specific version.
Then you just add one entry for the other project. And assign the respective version to the projects.

It's better to be explicit than implicit. 

If the application in settings state "Houdini 20.5.332" then I'd be quite pissed as admin that it'd secretly run another version on some machine - I want _that_ version!

> i believe the setting for linux should feel natural for the linux user, the same for windows and mac.

Do note that, it's not for the user - but for the admin.

---

@m-u-r-p-h-y @antirotor any thoughts?

Then we can pick the one we want for Linux here and merge this.

antirotor:
The more specific one. When you configure as an admin version of DCC, it is usually the  specific version because it is tested and your plug-in works with it. There might be reason to use different version on Windows than Linux because of some particular compatibility issue but that should be always considered some edge case and that change must be done willingly by the admin himself.

Lypsolon (approved):
This is open for a long time so Im just gonna say. 
Im Good with the changes and i think its good to go.

server/applications.json
```diff
@@ -648,8 +648,10 @@
...
+                        ],
+                        "linux": ["/opt/hfs19.5.805/bin/houdini"]
```
MustafaJafar, line 654: I think it hasn't `/bin`. 
I don't use the app on linux but I saw this in a [YT tutorial](https://www.youtube.com/watch?v=ZPR5lp7GSkc). 
image

```suggestion
                        "linux": ["/opt/hfs19.5.805/houdini"]
```

BigRoy: Some client communication is showing that there is indeed a `/bin` subfolder.

[private thread](https://discord.com/channels/517362899170230292/1268146246917296232/1268892825529876634)

I think is this particular video it works from that folder because he runs `source houdini_setup` which initializes the houdini env and _likely_ adds the `bin/` folder to `PATH` or alike.

I'm quite confident the paths are correct as they are now :D

Lypsolon: jap this works because of the source Houdini setup. 
it allows you to run Houdini from where ever you want.

PR #20, opened 2024-08-28
Add section names to settings

MustafaJafar:
## Changelog Description
This PR provides extra clarity about different sections in the addon settings by add section names for settings.
This will be helpful for writing the documentation.

image

## Testing notes:
Everything should work as usual

### DIFF
--- a/server/settings.py
+++ b/server/settings.py
@@ -433,7 +433,8 @@ class ApplicationsAddonSettings(BaseSettingsModel):
     applications: ApplicationsSettings = SettingsField(
         default_factory=ApplicationsSettings,
         title="Applications Definitions",
-        scope=["studio"]
+        scope=["studio"],
+        section="Definitions"
     )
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
For context. Here's a _before_ screenshot.

image

I'm **not completely convinced** it's needed but it does look clearer at first sight so fine by me.

Maybe @Innders or @mkolar has better ideas?

BigRoy:
> I can understand the motivation behind it but I also agree with @BigRoy. I had to inspect the code to actually spot the difference.

Does that mean you approve or disapprove?

MustafaJafar:
> I'm **not completely convinced** it's needed but it does look clearer at first sight so fine by me.

It's not needed. but it looks nicer in the docs
image

BigRoy:
I'd say - merge this. It's cosmetics, and doesn't affect settings much if we redesign later?

PR #24, opened 2024-09-06
Chore: Update Application variants in default settings

iLLiCiTiT:
## Changelog Description
Removed old variants and added more recent releases.

## Testing notes:
1. Validate added variants.
2. Make sure the defaults are actually working.

### DIFF
--- a/server/applications.json
+++ b/server/applications.json
@@ -866,7 +866,9 @@
                         "windows": [
                             "C:\\Program Files\\Adobe\\Adobe Photoshop 2023\\Photoshop.exe"
                         ],
-                        "darwin": [],
+                        "darwin": [
+                            "/Applications/Adobe Photoshop 2023/Adobe Photoshop 2023.app"
+                        ],
                         "linux": []
                     },
                     "arguments": {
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Looks good to me. 

However, we still get plenty of users using Maya 2022 (which is Py3) so I'd personally keep that one in for the time being.

m-u-r-p-h-y:
How difficult would it be to eliminate the inconsistency in the naming of apps in Application addon settings?

`Autodesk <APP>`
` Adobe <APP> `

and rest is just simple:
Nuke, Blender, Resolve, Houdini, etc.

I suppose it is only a label so it should not affect apps in anatomy templates in any way . . .

iLLiCiTiT:
You simply should update to this PR 
image

iLLiCiTiT:
> I copied the applications.json to my server, restarted and reset to defaults to test . . .

Labels are not defined in `applications.json` .

MustafaJafar:
> However, we still get plenty of users using Maya 2022 (which is Py3) so I'd personally keep that one in for the time being.

It's also worth mentioning that Maya 2022 may not be compatible with some addons. 
like this issue in kitsu: https://community.ynput.io/t/i-get-an-error-when-i-use-mayas-ayon-publishing-tool-how-can-i-fix-it/1677

iLLiCiTiT:
> However, we still get plenty of users using Maya 2022 (which is Py3) so I'd personally keep that one in for the time being.

Is that reason to have it in defaults? You can still add it if you need it.

BigRoy:
> Is that reason to have it in defaults? You can still add it if you need it.

I'd say if many clients are using it then - yes, it's a valid default. But happy to stand corrected by @m-u-r-p-h-y or @LiborBatek

LiborBatek:
@BigRoy I think removing ```maya 2022``` should be the way imho ...as been said "can be added manually"

I guess this maya version is no longer fully supported by default, or rather not the one users should stick to (by default)? I could be wrong tho...

Issue #26, opened 2024-09-16
Web actions show all applications regardless of their state


antirotor:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

No matter what applications I have enabled in the applications addon settings, web actions shows all of them. Observe:

![image](https://github.com/user-attachments/assets/3ab38641-f5da-42e2-b7c0-5186f9b8f62b)
![image](https://github.com/user-attachments/assets/b2aa28c8-0acc-4fc4-b10d-543593ec4517)

with this, I should not be able to see Photoshop, Substance Painter and others.

### Expected Behavior:

Only enabled applications should be shown. This should ofc respect individual bundles.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

- disable some applications in applications addon settings
- compare it with the list of applications shown on task

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

PR #33, opened 2024-09-19
Webactions: Skip disabled applications

iLLiCiTiT:
## Changelog Description
Disabled applications are not shown in webactions.

## Additional info
When profile specified that all applications should be visible based on profile filters, it showed even disabled applications.

## Testing notes:
1. Create package, upload it to AYON and add it to bundle.
2. Create profile in applications profiles to show all applications.
3. Only enabled applications should be shown on selected task.

Resolves https://github.com/ynput/ayon-applications/issues/26


64qam:
Doesnt work for me, it always show all applications or nothing.

64qam (approved):

PR #34, opened 2024-09-19
Default settings: VRay Maya has maya host

iLLiCiTiT:
## Changelog Description
Example VRay Maya tool have invalid hosts filter.

## Additional info
There was invalid filter for hosts with `""` in host names filter.

## Testing notes:
1. Validate changes.


m-u-r-p-h-y (approved):
even though it is not required to specify the host and can be used without it,, we can use it as an example. 

approved

Issue #41, opened 2024-12-09
Fill environment variables for farm


iLLiCiTiT:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Use `FARM_JOB_ENV_DATA_KEY` from ayon core to fill environment variables necessary for farm publishing.
- `AYON_APP_NAME`

Will require to change required ayon-core to `>=1.0.10`.

PR #43, opened 2025-01-03
Fill environment variables for farm jobs

iLLiCiTiT:
## Changelog Description
Fill environment variables for farm in publish plugins.

Move control to collect applications environment variables, needed for farm jobs, to applications addon.

## Testing notes:
1. At this moment probably nothing changed because deadline has to stop auto-fill the key.

Resolves https://github.com/ynput/ayon-applications/issues/41

### DIFF
--- a/client/ayon_applications/plugins/publish/collect_farm_env_variables.py
+++ b/client/ayon_applications/plugins/publish/collect_farm_env_variables.py
@@ -1,0 +1,26 @@
+import os
+
+import pyblish.api
+
+try:
+    from ayon_core.pipeline.publish import FARM_JOB_ENV_DATA_KEY
+except ImportError:
+    # NOTE Can be removed when ayon-core >= 1.0.10 is required in package.py
+    FARM_JOB_ENV_DATA_KEY = "farmJobEnv"
+
+
+class CollectApplicationsJobEnvVars(pyblish.api.ContextPlugin):
+    """Collect set of environment variables to submit with deadline jobs"""
+    order = pyblish.api.CollectorOrder - 0.45
+    label = "Collect Applications farm environment variables"
+    targets = ["local"]
+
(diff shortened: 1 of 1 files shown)


kalisp:
Is this moved here (`application`) as it is just not only for deadline, but for possible other farm controllers? (Eg. it is not in `ayon-deadline` where I would expect that).

If thats the case I would just remove last mention of `deadline` in docstrings.

client/ayon_applications/plugins/publish/collect_farm_env_variables.py
```diff
@@ -0,0 +1,26 @@
...
+
+try:
+    from ayon_core.pipeline.publish import FARM_JOB_ENV_DATA_KEY
+except ImportError:
+    # NOTE Can be removed when ayon-core >= 1.0.10 is required in package.py
+    FARM_JOB_ENV_DATA_KEY = "farmJobEnv"
+
+
+class CollectApplicationsJobEnvVars(pyblish.api.ContextPlugin):
+    """Collect set of environment variables to submit with deadline jobs"""
```
iLLiCiTiT, line 13:
```suggestion
    """Collect set of environment variables for farm jobs"""
```

kalisp (approved):
I am guessing that creating this plugin here (and not in ayon-core) its because there is already `collect_app_name` so it kinda make sense to have this at same place. 

(Had to wrap my head around what is benefit, so just writing it down for future generations ;)

iLLiCiTiT:
> I am guessing that creating this plugin here (and not in ayon-core) its because there is already collect_app_name so it kinda make sense to have this at same place.
>
> (Had to wrap my head around what is benefit, so just writing it down for future generations ;)

Because ayon-core does not define it and even should not know about the env variable.

PR #49, opened 2025-01-15
Adds Gaffer support

sosoyan:
## Changelog Description
Added support for Gaffer and should work with ayon-gaffer addon 
https://github.com/RVXStudio/ayon-gaffer

https://github.com/GafferHQ/gaffer

## Testing notes:
1. Add a path to your gaffer executable in Application Settings
2. Run as any other application from Launcher


server/applications.json
```diff
@@ -173,6 +173,23 @@
...
+            "host_name": "gaffer",
+            "environment": "{}",
+            "variants": [
+                {
+                    "name": "1.5.2",
+                    "label": "1.5.2",
+                    "executables": {
+                        "windows": [],
+                        "darwin": [],
+                        "linux": []
```
BigRoy, line 187: Although likely "incorrect" in the majority of cases I wonder if we should default these paths to what makes sense according to the Gaffer install guide: https://www.gafferhq.org/documentation/1.5.2.0/GettingStarted/InstallingGaffer/index.html

In particular:
- Windows: `C:\software\gaffer-1.5.2.0-windows`
- Linux: `/opt/gaffer-1.5.2.0-linux`

Mostly because by adding _something_ there's at least an example of _what_ binary to point to. Is it a `.bat` - or another? etc.

Could you add that @sosoyan ?

BigRoy (approved):
Looks good to me - I think it'd be nice to add default paths for windows and linux as pointed out in [my comment](https://github.com/ynput/ayon-applications/pull/49/files#r1918303828) because it'd aid admins to know what to point to.

Other than that, this is indeed what is needed to add gaffer app support.

Might be nice in the PR description to point to a matching **ayon-gaffer** integration addon repository that may work with this PR - just so it's tracked here for the future.

E.g. I know of https://github.com/RVXStudio/ayon-gaffer

server/applications.json
sosoyan: Sure will add both

MustafaJafar (approved):
It works. 
Because there's no addon for it I think AYON will only launch it. 
image

sosoyan:
> It works. Because there's no addon for it I think AYON will only launch it. image

I'm working on minimal ayon-gaffer addon will pr when it's ready.

sosoyan:
> I'm working on minimal ayon-gaffer addon will pr when it's ready.

I've pushed a minimal host addon now 
https://github.com/sosoyan/ayon-gaffer

sosoyan:
What holds this to be merged?

BigRoy:
So far it's been mostly internal discussion whether we want to merge this already prior to an addon being available on the addon marketplace directly. That was mostly it. At the same time, it being there isn't much in the way. Also, for studios rolling out their own addons managing this PR's branch into their own branch also isn't necessarily the big amount of effort - which also made it high priority to make a decision right away. Anyway, will raise it again.

@philippe-ynput

philippe-ynput (approved):
Good for merging. Thanks @sosoyan !

PR #50, opened 2025-01-16
Add substance designer as applications

moonyuet:
## Changelog Description
Add substance designer as default applications

## Additional review information
Use together with pre-launch hook https://github.com/ynput/ayon-core/pull/1089 and initial integration.

## Testing notes:
1. Build and Install addon
2. There should be substance designer as default addon

PR #52, opened 2025-01-22
Substance designer: Define label and icon on server

iLLiCiTiT:
## Changelog Description
Server has missing icon and label definitions, it is defined only in client (for AYON launcher).

## Testing notes:
1. Webactions show correct label and icon.

Fix related to https://github.com/ynput/ayon-applications/pull/50


BigRoy (approved):
Should be the correct fix. 👍

moonyuet (approved):

PR #53, opened 2025-02-03
Remove 3dsmax addon directory enviroment

moonyuet:
## Changelog Description
Remove 3dsmax addon directory environment as default setting regarding to https://github.com/ynput/ayon-3dsmax/pull/34

n/

## Testing notes:
1. Build and install application addon alongside the Max addon built with https://github.com/ynput/ayon-3dsmax/pull/34
2. Launch Max
3. AYON plugins should be installed and working successfully.

### DIFF
--- a/server/applications.json
+++ b/server/applications.json
@@ -73,7 +73,7 @@
         "adsk_3dsmax": {
             "enabled": true,
             "host_name": "max",
-            "environment": "",
+            "environment": "{}",
             "variants": [
                 {
                     "name": "2024",
--- a/server/applications.json
+++ b/server/applications.json
@@ -73,7 +73,7 @@
         "adsk_3dsmax": {
             "enabled": true,
             "host_name": "max",
-            "environment": "{\n    \"ADSK_3DSMAX_STARTUPSCRIPTS_ADDON_DIR\": \"{OPENPYPE_ROOT}/openpype/hosts/max/startup\"\n}",
+            "environment": "",
             "variants": [
                 {
                     "name": "2024",


LiborBatek (approved):
I have followed the testing steps and all seems working normally and without issues.

LGTM

BigRoy:
This should come with a version check requirement between application addon and 3dsmax addon

By removing this here, and adding it to 3dsmax I believe the applications addon must be targeted to 'require' the newer 3dsmax addon version when used together.

@iLLiCiTiT thoughts?

server/applications.json
```diff
@@ -73,7 +73,7 @@
         "adsk_3dsmax": {
             "enabled": true,
             "host_name": "max",
-            "environment": "{\n    \"ADSK_3DSMAX_STARTUPSCRIPTS_ADDON_DIR\": \"{OPENPYPE_ROOT}/openpype/hosts/max/startup\"\n}",
+            "environment": "",
```
BigRoy, line 76: Shouldn't the default be `"{}"`?
```suggestion
            "environment": "{}",
```

iLLiCiTiT: Yes it should

iLLiCiTiT:
> This should come with a version check requirement between application addon and 3dsmax addon

Question of the centurie. The decisions made in past was that 3dsmax addon should require higher ayon-applications, and ayon-applications should not care.

But, the environment variable set here would not work. 1. It uses `OPENPYPE_ROOT` which is not set anymore. 2. It points to `{OPENPYPE_ROOT}/openpype/hosts/max/startup` which does not exist. So the question is how id did work?

iLLiCiTiT (approved):

PR #57, opened 2025-03-19
Chore: Use env tools functions from ayon core

iLLiCiTiT:
## Changelog Description
Use `merge_env_variables` and `compute_env_variables_structure` from ayon-core repository if are available.

The fallbacks should be removed after some time, which will require bump compatible version of ayon-core.

## Testing notes:
1. Test with latest ayon-core.
2. Make sure your task is using tools that have env variables to resolve.
3. The environments of application and tools should be resolved as before.

### DIFF
--- a/client/ayon_applications/utils.py
+++ b/client/ayon_applications/utils.py
@@ -31,6 +31,7 @@ from .manager import ApplicationManager
 # NOTE Remove try -> except block, when ayon-core > 1.1.1 is required,
 #   or require the version to remove it.
 try:
+    # Was introduced in ayon-core 1.1.2
     from ayon_core.lib import (
         merge_env_variables,
         compute_env_variables_structure,
--- a/client/ayon_applications/utils.py
+++ b/client/ayon_applications/utils.py
@@ -4,7 +4,6 @@ import json
 import platform
 import collections
 
-import acre
 import ayon_api
 
 from ayon_core import AYON_CORE_ROOT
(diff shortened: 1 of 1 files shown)


kalisp (approved):
Looks reasonable, but I am not sure how to test it. I tried to play with `ayon+settings://applications/project_tools` but couldnt notice if it gets actually triggered.

client/ayon_applications/utils.py
```diff
@@ -28,6 +28,19 @@
...
+#   or require the version to remove it.
+try:
+    # Was introduced in ayon-core 1.1.2
+    from ayon_core.lib import (
+        merge_env_variables,
+        compute_env_variables_structure,
+    )
+except ImportError:
+    merge_env_variables = None
+    compute_env_variables_structure = None
```
BigRoy, line 41: Why not move the "legacy function definitions" here?
So that in the code we're always using the new naming already - and at some point we'd just need to remove the `ImportError` fallback?

So in the except case we'd just do:

```python
def merge_env_variables(env, current_env):
    """Modified function(merge) from acre module."""
    result = current_env.copy()
    for key, value in env.items():
        # Keep missing keys by not filling `missing` kwarg
        value = acre.lib.partial_format(value, data=current_env)
        result[key] = value
    return result

def compute_env_variables_structure(merged_env):
    return acre.compute(merged_env, cleanup=False)
```

kalisp: Ohh, lovely, I liek it :).

iLLiCiTiT: Done

BigRoy (approved):
Didn't test it, but changes look good to me

PR #64, opened 2025-04-02
Server icons: Add missing server icon mapping

iLLiCiTiT:
## Changelog Description
Added missing icon mappings.

Fix icons showed in webactions.

## Testing notes:
1. All applications will have correct icon in webactions.


MilaKudr:
The full list of applications is fine, but there is another list, probably applications that are available on the machine, and this shows wrong icons....
Screenshot 2025-04-03 095356

iLLiCiTiT:
> The full list of applications is fine, but there is another list, probably applications that are available on the machine, and this shows wrong icons....

Those actions are from example addon, cannot be affected by this addon at all.

MilaKudr (approved):

PR #76, opened 2025-05-22
Chore: Store config per task

iLLiCiTiT:
## Changelog Description
Fix config storing per task.

Addon defined wrong method name to create config hash, fixed method name and added more small modifications.

## Testing notes:
1. Changing of 'Skip last workfile' is stored per task.


BigRoy (commented):
This seems to work, values are now per task.

I do have a bug that if I set "Do not open last workfile" for an Application Group (e.g. combined Maya 2025/2026 entries) then it only skips opening workfile for the top entry (2026 in this case).

If I then disable Maya 2026 application, and add 2024 so that I have app group 2025/2024 then suddenly it's not marked to skip last workfile anymore. So I right click and enable it again.

If I re-enable Maya 2026 - then that one is still overidden to not open last workfile.
In short, the value is now stored per application version (and there's no way to toggle it for entries in an application group that are NOT the last version).

I think instead, the toggle should act/store on the Application Group, instead of a specific version/variant.

iLLiCiTiT:
> I do have a bug that if I set "Do not open last workfile" for an Application Group (e.g. combined Maya 2025/2026 entries) then it only skips opening workfile for the top entry (2026 in this case).

Reproduced and fixed in https://github.com/ynput/ayon-core/pull/1226 .

BigRoy (approved):
> > I do have a bug that if I set "Do not open last workfile" for an Application Group (e.g. combined Maya 2025/2026 entries) then it only skips opening workfile for the top entry (2026 in this case).
> 
> Reproduced and fixed in [ynput/ayon-core#1226](https://github.com/ynput/ayon-core/pull/1226) .

Fixed in the other PR now. Looks good.

server/__init__.py
```diff
@@ -156,24 +157,27 @@ async def pre_setup(self):
...
-        if identifier.startswith(IDENTIFIER_PREFIX):
-            # Change identifier to only app name
-            identifier = identifier[len(IDENTIFIER_PREFIX):]
+        if not identifier.startswith(IDENTIFIER_PREFIX):
+            return await super().create_action_config_hash(
+                identifier, context, user, variant
+            )
+
+        # Change identifier to only app name and one task id
+        identifier = identifier[len(IDENTIFIER_PREFIX):]
```
BigRoy, line 174: Since the server runs a recent python version - did you know you could do `identifier.removeprefix(IDENTIFIER_PREFIX)`?

iLLiCiTiT: Good point, will do.

64qam:
Works on Windows, Linux, Mac +Nuke 15, doesn't work on Windows + Nuke 13.

PR #81, opened 2025-06-20
Speedtree icon: Use standard PNG format

iLLiCiTiT:
## Changelog Description
Replaced icon for speedtree application. Used 512x512 icon.

Qt didn't like format of PNG used for speedtree icon.
```terminal
qt.gui.imageio: libpng warning: iCCP: known incorrect sRGB profile
```

## Testing notes:
1. Upload addon to server.
2. Showing the application in launcher tool won't cause warning `qt.gui.imageio: libpng warning: iCCP: known incorrect sRGB profile`


BigRoy (approved):
After creating package and uploading to server, then restart tray launcher I indeed don't get this warning anymore. So seems to work!

I'm a bit confused that I needed to upload to the server for this to work since I assumed the client side app, and the icons it showed were served _from the client_ and hence would be shown from the dev mode loaded codebase?

Anyway, works!

iLLiCiTiT:
> I'm a bit confused that I needed to upload to the server for this to work since I assumed the client side app, and the icons it showed were served from the client and hence would be shown from the dev mode loaded codebase?

With current core develop are icons loaded only from server.

PR #84, opened 2025-06-27
Add settings for Marmoset Toolbag 5 to AYON

muchenhen:
## Changelog Description
I have added the application information for Marmoset

We need to use this software in the process and are currently developing the corresponding addon. We believe that Marmoset is still widely used in some industries, and since it cannot display icons only when configured on the web, we suggest adding it directly to applications

## Testing notes:

1. Install Marmoset Toolbag 5
3. Check if the corresponding icon appears in AYON's actions
4. 3. Start Marmoset in the launcher and check if it started successfully


muchenhen:
Sorry, I didn't carefully review the workflow. I will close this manually created PR first

PR #94, opened 2025-08-07
Store process information and manage application logging

antirotor:
## Changelog Description
Store launched process information in a local database and add process monitoring/output manager.

When process is run, this will store information about it in local sqlite3 database that can be later on used by other parts of AYON. It also passes temporary file as *stdout* and *stderr* handle, storing its path along the process information.

All this information can be used to read out what the launched process is outputting in situations where it doesn't open its own console. For this Process Monitoring tool is added:

image

This tool can be run using `Process Monitor` item in Tray menu.

**What it does:**
- periodically refresh process list from database
- checks if process is still running
- can open file with process output and reload it periodically
- you can clean output files for any inactive process
- it colorize output if it contains ANSI sequences
- it can handle wrapping shell scripts as executables (scripts that will at the end launch the "real" executable)

## Further enhancements:
- connect to message broker and send logs there (https://github.com/ynput/ayon-rabbitmq)

## Shell scripts and passing of PID back to AYON
When launching applications through shell scripts on Linux, the process ID returned by the `app_launcher` (mid-process) corresponds to the shell process, not the actual application process that the shell script launches.

### Architecture Overview
- **AYON Applications Manager** (`manager.py`) creates launch contexts and manages application processes
- **App Launcher** (`app_launcher.py`/`app_launcher.cpp` in [ayon-launcher](https://github.com/ynput/ayon-launcher)) is the mid-process that handles detached process launching on Linux
- **Shell Scripts** are often used as executables that set up environment and launch actual applications

### Solution
The solution implements a **PID file convention** that allows shell scripts to communicate the actual application PID back to AYON through the app_launcher mid-process.

### How it Works

1. **PID File Creation**: The AYON manager creates a temporary PID file and provides its path to the app_launcher via JSON data
2. **Environment Variable**: The app_launcher passes the PID file path to shell scripts via `AYON_PID_FILE` environment variable
3. **Shell Script Responsibility**: Shell scripts can write the actual application PID to this file after launching the application
4. **PID Discovery**: The app_launcher waits briefly and checks for an updated PID in the file, using it if found

### Integration Requirements

For shell scripts to take advantage of this feature:
1. Check if `$AYON_PID_FILE` environment variable exists
2. Launch your application in the background using `&`
3. Capture the PID using `$!` (bash) or equivalent
4. Write the PID to the file specified by `$AYON_PID_FILE`

#### Bash Example
```bash
#!/bin/bash

# Launch the actual application in background
/path/to/actual/application "$@" &

# Get the PID of the actual application
APP_PID=$!

# Write the PID to the AYON PID file if available
if [ -n "$AYON_PID_FILE" ]; then
    echo "$APP_PID" > "$AYON_PID_FILE"
fi

# Wait for the application to finish (optional)
wait $APP_PID
```
---

> [!WARNING]
For retrieving process status, it is using `psutil` tool that you need in your dependencies. That means building new dependency package. You can add `psutil` manually but keep in mind that it is binary dependency and thus it will work only with compatible Python host (if you build it for python 3.9 it is unlikely that it will also work in 3.11).

## Testing notes:
You should be able to run any DCC as usual and see its output in **Process Manager**. It works on Linux and macos. Try to wrap and execute some application in shell script as shown in the example above - **Process Manager** should still pick the correct process id and monitor the process. Removing of selected item/All inactive items should work too. If you use ANSI sequences to colorize your shell script, you should see that also in the output (it supports even bold, underline, ... styles).

### DIFF
--- a/client/ayon_applications/version.py
+++ b/client/ayon_applications/version.py
@@ -1,3 +1,3 @@
 # -*- coding: utf-8 -*-
 """Package declaring AYON addon 'applications' version."""
-__version__ = "1.2.1+dev"
+__version__ = "1.2.2+dev"
--- a/package.py
+++ b/package.py
@@ -1,6 +1,6 @@
 name = "applications"
 title = "Applications"
-version = "1.2.1+dev"
+version = "1.2.2+dev"
 
 client_dir = "ayon_applications"
 
--- a/server/applications.json
+++ b/server/applications.json
@@ -876,7 +876,7 @@
         "harmony": {
             "enabled": true,
             "host_name": "harmony",
-            "environment": "{\n    \"AYON_HARMONY_WORKFILES_ON_LAUNCH\": \"1\"\n}",
+            "environment": "{}",
             "variants": [
                 {
                     "name": "22",
@@ -962,7 +962,7 @@
         "photoshop": {
             "enabled": true,
             "host_name": "photoshop",
-            "environment": "{\n    \"AVALON_PHOTOSHOP_WORKFILES_ON_LAUNCH\": \"1\",\n    \"WORKFILES_SAVE_AS\": \"Yes\"\n}",
+            "environment": "{\n  \"WORKFILES_SAVE_AS\": \"Yes\"\n}",
             "variants": [
                 {
                     "name": "2023",
@@ -1026,7 +1026,7 @@
         "aftereffects": {
             "enabled": true,
             "host_name": "aftereffects",
-            "environment": "{\n    \"AVALON_AFTEREFFECTS_WORKFILES_ON_LAUNCH\": \"1\",\n    \"WORKFILES_SAVE_AS\": \"Yes\"\n}",
+            "environment": "{\n    \"WORKFILES_SAVE_AS\": \"Yes\"\n}",
             "variants": [
                 {
                     "name": "2023",
@@ -1090,7 +1090,7 @@
         "premiere": {
             "enabled": true,
             "host_name": "premiere",
-            "environment": "{\n    \"AYON_PREMIERE_WORKFILES_ON_LAUNCH\": \"1\",\n    \"WORKFILES_SAVE_AS\": \"Yes\"\n}",
+            "environment": "{\n    \"WORKFILES_SAVE_AS\": \"Yes\"\n}",
             "variants": [
                 {
                     "name": "2023",


client/ayon_applications/manager.py
```diff
@@ -163,12 +359,167 @@ def launch(self, app_name, **data):
...
+                site_id=row[8],
+            )
+            for row in rows
+        )
+        # Check if processes are still running
+        pids = [proc.pid for proc in processes if proc.pid is not None]
+        if pids:
+            running_status = self._are_processes_running(pids)
+            for proc, is_running in zip(processes, running_status):
+                proc.active = is_running[1]
```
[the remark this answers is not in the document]
antirotor: Doing this individually is very slow for the fallback options where you can batch query processes - at least on Windows. If the fallback isn't needed (and that remains to be seen), we might change it to check individual processes/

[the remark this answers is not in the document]
antirotor: Added some info about this to docstrings.

client/ayon_applications/manager.py
```diff
@@ -1,11 +1,17 @@
...
 import os
 import sys
 import copy
+from hashlib import sha256
 import json
+from pathlib import Path
 import tempfile
 import platform
 import inspect
+import sqlite3
```
iLLiCiTiT, line 12: Are we sure this import won't break something in DCCs?

client/ayon_applications/manager.py
```diff
@@ -47,6 +89,160 @@ def __init__(self, studio_settings=None):
...
+            "CREATE TABLE IF NOT EXISTS process_info ("
+            "hash TEXT PRIMARY KEY, "
+            "name TEXT, "
+            "args TEXT DEFAULT NULL, "
+            "env TEXT DEFAULT NULL, "
+            "cwd TEXT DEFAULT NULL, "
+            "pid INTEGER DEFAULT NULL, "
+            "output_file TEXT DEFAULT NULL, "
+            "created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP, "
+            "site_id TEXT DEFAULT NULL"
```
iLLiCiTiT, line 128: Why is site id used? The database is located at the same place where site id is defined...

client/ayon_applications/manager.py
```diff
@@ -163,12 +359,188 @@ def launch(self, app_name, **data):
...
+        try:
+            import psutil
+            for pid in pids:
+                is_running = psutil.pid_exists(pid)
+                result.append((pid, is_running))
+
+            return result
+
+        except ImportError:
+            # Fallback for systems without psutil
```
iLLiCiTiT, line 515: The code should not live under the `ImportError` handler.
```suggestion
        except ImportError:
            pass

        # Fallback for systems without psutil
```

client/ayon_applications/manager.py
```diff
@@ -163,12 +359,188 @@ def launch(self, app_name, **data):
...
+                        ["tasklist", *filters], capture_output=True, text=True
+                    )
+                    for line in tasklist_result.stdout.splitlines():
+                        for pid in pids:
+                            if f"{pid}" in line:
+                                result.append((pid, True))
+                                break
+                except subprocess.SubprocessError:
+                    return []
+            else:
```
iLLiCiTiT, line 532: No need for else statement as windows condition always returns result.

client/ayon_applications/manager.py
```diff
@@ -537,27 +933,85 @@ def _run_process(self):
...
+            prefix=f"ayon_{self.application.host_name}_output_",
+            suffix=".txt",
+            delete=False
+        )
+        temp_file_path = temp_file.name
+        temp_file.close()
+        with open(temp_file_path, "wb") as tmp_file:
+            self.kwargs["stdout"] = tmp_file
+            self.kwargs["stderr"] = tmp_file
+            process = subprocess.Popen(self.launch_args, **self.kwargs)
```
iLLiCiTiT, line 999: This won't create detached process on windows.

client/ayon_applications/ui/process_monitor.py
```diff
@@ -0,0 +1,697 @@
...
+from typing import TYPE_CHECKING, Optional, Union
+
+from qtpy import QtWidgets, QtCore, QtGui
+from qtpy.QtCore import QRunnable, Slot, QThreadPool
+from qtpy.QtCore import QModelIndex, QPersistentModelIndex
+
+from ayon_core.style import load_stylesheet
+from ayon_core.tools.utils import get_ayon_qt_app
+
+from ..manager import ApplicationManager
```
iLLiCiTiT, line 16:
```suggestion
from ayon_applications.manager import ApplicationManager
```

client/ayon_applications/ui/process_monitor.py
```diff
@@ -0,0 +1,697 @@
...
+        """Refresh process data in background thread."""
+        with CatchTime() as timer:
+            try:
+                processes = self._manager.get_all_process_info()
+                self.signals.finished.emit(processes)
+            except Exception as e:
+                self.signals.error.emit(str(e))
+        self._log.debug(
+            "Refresh from db completed in %s", timer.readout)
+
```
iLLiCiTiT, line 64:
```suggestion


```

client/ayon_applications/ui/process_monitor.py
```diff
@@ -0,0 +1,697 @@
...
+        except Exception as e:
+            self.signals.error.emit(f"Error reading file: {e}")
+
+
+class CleanupWorkerSignals(QtCore.QObject):
+    """Signals for CleanupWorker."""
+    # Emits (deleted_processes, deleted_files)
+    finished = QtCore.Signal(int, int)
+    error = QtCore.Signal(str)
+
```
iLLiCiTiT, line 101:
```suggestion


```

client/ayon_applications/ui/process_monitor.py
```diff
@@ -0,0 +1,697 @@
...
+                os.remove(target_process.output)
+                file_deleted = 1
+        # Delete from database
+        deleted = self._manager.delete_process_info(self._process_hash)
+        process_deleted = 1 if deleted else 0
+
+        self.signals.finished.emit(process_deleted, file_deleted)
+
+
+class ProcessTableModel(QtCore.QAbstractTableModel):
```
iLLiCiTiT, line 191: Oh boy, you're making you life harder. You can use `QStandardItemModel` that implements most of what you do here and would probably make it faster.

client/ayon_applications/ui/process_monitor.py
iLLiCiTiT, line 1: Review of this file would take hours. I can ignore it, or ask for permission to rewrite it...

client/ayon_applications/ui/process_monitor.py
```diff
@@ -0,0 +1,697 @@
...
+    """Worker thread for loading file content."""
+
+    def __init__(self, file_path):
+        super().__init__()
+        self.signals = FileContentWorkerSignals()
+        self.signature = self.__class__.__name__
+        self._file_path = file_path
+        self._log = getLogger(self.signature)
+
+    @Slot()
```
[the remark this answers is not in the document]
antirotor: For threading and signals, it is better (and little bit faster).

client/ayon_applications/ui/process_monitor.py
antirotor: Maybe just point out what's wrong in general way if you think so.

client/ayon_applications/ui/process_monitor.py
antirotor: I am not really sure about that. Maybe just `headerData()` but rest of it is custom anyway. More importantly - it works and there is nothing inherently wrong about that, or is it?

client/ayon_applications/manager.py
```diff
@@ -163,12 +359,188 @@ def launch(self, app_name, **data):
...
+        # This is done by checking the pid of the process.
+        # It is using `_are_processes_running` method which that
+        # checks for processes in batch, mostly because of the fallback
+        # on systems without `psutil` module. See `_are_processes_running`
+        # documentation for more details.
+        pids = [proc.pid for proc in processes if proc.pid is not None]
+        if pids:
+            running_status = self._are_processes_running(pids)
+            for proc, is_running in zip(processes, running_status):
+                proc.active = is_running[1]
```
iLLiCiTiT, line 405:
```suggestion
            for proc, (_, is_running) in zip(processes, running_status):
                proc.active = is_running
```

client/ayon_applications/ui/process_monitor.py
```diff
@@ -0,0 +1,697 @@
...
+        toolbar_layout.addWidget(self._loading_label)
+
+        main_layout.addLayout(toolbar_layout)
+
+        # Splitter for table and output
+        splitter = QtWidgets.QSplitter(QtCore.Qt.Orientation.Vertical)
+
+        # Process table
+        self._table_model = ProcessTableModel()
+        self._table_view = QtWidgets.QTableView()
```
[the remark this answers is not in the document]
iLLiCiTiT: Tree view is row based instead of cell based.

- List view is row based, but supports only 1 column.
- Tree view is row based, multiple columns, items can have children (don't have to).
- Table view is cell based -> selection is handled for combination of row and column -> causing wierd look, usually have borders around each cell.

client/ayon_applications/manager.py
iLLiCiTiT: If you launch Nuke or Photoshop, it is separated and the stdout and stderr are not outputed to AYON terminal?

client/ayon_applications/ui/process_monitor.py
```diff
@@ -0,0 +1,735 @@
...
+from typing import Optional, Union
+
+from qtpy import QtWidgets, QtCore, QtGui
+from qtpy.QtCore import QRunnable, Slot, QThreadPool
+from qtpy.QtCore import QModelIndex, QPersistentModelIndex
+
+from ayon_core.style import load_stylesheet
+from ayon_core.tools.utils import get_ayon_qt_app
+
+from ..manager import ApplicationManager, ProcessInfo
```
iLLiCiTiT, line 16:
```suggestion
from ayon_applications.manager import ApplicationManager, ProcessInfo
```

client/ayon_applications/ui/process_monitor.py
iLLiCiTiT: Firs of all formatting.
In general to keep layouts correct and propagate styles it is better to always pass parent to a widget on it's initialization (`.setParent` does not work without repolishing the widget). The same with layout.

It is better to explicitly set layout content margins because each platform has different defaults.

This is my preference: Create all widgets and objects and then store them to `self` at the end. Usually makes it a lot easier to read the code and also allows to have 1 place where to look what variables are actually stored to `self`.

Do not recreate window or timer or thread. Create widget/window once, or make sure the previous is closed, hidden and destroyed, the same with timers and threads.

I think you can use `QThread` instead of `QRunnable` it allows to use signals and is in fact the same. It is still python code you're running there so it is not really a thread (GIL).

client/pyproject.toml
```diff
@@ -0,0 +1,8 @@
+name="applications"
+description="AYON Applications addon"
+
+[tool.poetry.dependencies]
+python = ">=3.9.1,<3.10"
+
```
iLLiCiTiT, line 6: Won't be used for anything.
```suggestion
```

client/ayon_applications/ui/process_monitor.py
```diff
@@ -0,0 +1,991 @@
...
+        """Replace current content with provided processes.
+
+        Args:
+            processes (list[ProcessInfo]): List of ProcessInfo objects.
+
+        """
+        self._processes = list(processes)
+        self.clear()
+        self.setColumnCount(len(self.headers))
+        self.setHorizontalHeaderLabels(self.headers)
```
iLLiCiTiT, line 317: `clear` resets the model to "initial state", not just rows but also headers and every other thing.
```suggestion
        root_item = self.invisibleRootItem()
        root_item.removeRows(0, root_item.rowCount())
```

iLLiCiTiT: Additional note:
It would be greate to NOT remove the items, but only update data of existing ones. By removing them you also loose any UI track of the process items (like selection).

client/ayon_applications/ui/process_monitor.py
```diff
@@ -0,0 +1,991 @@
...
+
+        # Loading indicator
+        self._loading_label = QtWidgets.QLabel("Loading...")
+        self._loading_label.setVisible(False)
+
+        toolbar_layout.addWidget(self._refresh_btn)
+        toolbar_layout.addWidget(self._clean_inactive_btn)
+        toolbar_layout.addWidget(self._clean_selected_btn)
+        toolbar_layout.addStretch()
+        toolbar_layout.addWidget(self._loading_label)
```
iLLiCiTiT, line 755:
```suggestion
        toolbar_layout.addWidget(self._refresh_btn, 0)
        toolbar_layout.addWidget(self._clean_inactive_btn, 0)
        toolbar_layout.addWidget(self._clean_selected_btn, 0)
        toolbar_layout.addStretch(1)
        toolbar_layout.addWidget(self._loading_label, 0)
```

client/ayon_applications/ui/process_monitor.py
```diff
@@ -0,0 +1,991 @@
...
+        # Auto-reload checkbox
+        self._auto_reload_checkbox = QtWidgets.QCheckBox(
+            "Auto-reload output for running processes (every 2s)")
+        self._auto_reload_checkbox.setChecked(True)
+        self._auto_reload_checkbox.toggled.connect(
+            self._on_auto_reload_toggled)
+
+        output_layout.addWidget(output_label)
+        output_layout.addWidget(self._output_text)
+        output_layout.addWidget(self._auto_reload_checkbox)
```
iLLiCiTiT, line 689:
```suggestion
        output_layout.addWidget(output_label, 0)
        output_layout.addWidget(self._output_text, 1)
        output_layout.addWidget(self._auto_reload_checkbox, 0)
```

client/ayon_applications/manager.py
```diff
@@ -46,6 +92,9 @@ class ApplicationManager:
             will always use these values. Gives ability to create manager
             using different settings.
     """
+    # holds connection to the process info storage
+    # - this is used to store process information about launched applications
+    _process_storage: Optional[sqlite3.Connection] = None
```
iLLiCiTiT, line 97: Cna we separate the whole process storage logic from `ApplicationManager`? It should really be another class. It is bloating the manager's logic a lot making it super complicated to follow which code is related to what.

antirotor:
This now makes use of https://github.com/ynput/ayon-launcher/pull/255

client/ayon_applications/addon.py
```diff
@@ -40,12 +42,40 @@
...
+                self._process_monitor_window is not None
+                and not self._process_monitor_window.isVisible()
+        ):
+            self._process_monitor_window.show()
+            self._process_monitor_window.raise_()
+            self._process_monitor_window.activateWindow()
+            return
+
+        self._process_monitor_window = ProcessMonitorWindow()
+        self._process_monitor_window.show()
```
iLLiCiTiT, line 76:
```suggestion
        if self._process_monitor_window is None:
            self._process_monitor_window = ProcessMonitorWindow()

        self._process_monitor_window.show()
        self._process_monitor_window.raise_()
```

client/ayon_applications/manager.py
```diff
@@ -524,26 +535,60 @@ def manager(self) -> ApplicationManager:
...
+            delete=False
+        ) as temp_file:
+            output_file = temp_file.name
+        # create temporary file to read back pid
+        with tempfile.NamedTemporaryFile(
+            mode="w", prefix="ayon_pid_", suffix=".txt", delete=False
+        ) as pid_temp_file:
+            pid_file = pid_temp_file.name
+
         json_data = {
```
iLLiCiTiT, line 582: What is `site_id` and `output` used for? Is there any reason to send `pid_file` to the subprocess?

client/ayon_applications/process.py
```diff
@@ -0,0 +1,682 @@
...
+    pid: Optional[int] = None
+    active: bool = False
+    output: Optional[Path] = None
+    start_time: Optional[float] = None
+    created_at: Optional[str] = None
+    site_id: Optional[str] = None
+
+
+class ProcessManager:
+    """Manager for handling processes in Ayon Applications."""
```
iLLiCiTiT, line 59:
```suggestion
    """Manager for handling processes in AYON Applications."""
```

client/ayon_applications/process.py
```diff
@@ -0,0 +1,682 @@
...
+                    exe_path = Path(exe_path_str)
+            except (
+                    psutil.NoSuchProcess,
+                    psutil.ZombieProcess,
+                    psutil.AccessDenied):
+                exe_path = None
+        return exe_path
+
+    @staticmethod
+    def _check_processes_running_win(
```
iLLiCiTiT, line 515: You can look at `_windows_pid_is_running` or rather `_windows_get_pid_args` in `ayon_core/tools/tray/lib.py`. Is using windows api to get the args to avoid running subprocess and parsing of the output.

client/ayon_applications/process.py
```diff
@@ -0,0 +1,682 @@
...
+
+    """
+
+    name: str
+    executable: Path
+    args: list[str]
+    env: dict[str, str]
+    cwd: str
+    pid: Optional[int] = None
+    active: bool = False
```
iLLiCiTiT, line 51: `active` means `running`?

client/ayon_applications/process.py
```diff
@@ -0,0 +1,682 @@
...
+    output: Optional[Path] = None
```
iLLiCiTiT, line 52: Is this the content of stdout/stderr files? Shouldn't that be loaded in a separate way (so you don't have to open the large files to list processes).

client/ayon_applications/process.py
```diff
@@ -0,0 +1,682 @@
...
+
+class ProcessIdTriplet(NamedTuple):
+    """Triplet of process identification values."""
+    pid: int
+    executable: str
+    start_time: Optional[float]  # the same goes for start time
+
+
+@dataclass
+class ProcessInfo:
```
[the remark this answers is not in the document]
iLLiCiTiT: And I don't understand why that was used over using uuid? The hash is potential danger from my perspective.

client/ayon_applications/process.py
```diff
@@ -0,0 +1,682 @@
...
+
+        Returns:
+            list[tuple[int, bool]]: List of tuples with process ID and
+                boolean indicating if the process is running.
+
+        """
+        result: list[tuple[int, bool]] = []
+        # Use tasklist CSV output for more robust parsing (handles spaces)
+        filters: list[str] = []
+        filters.extend(f"/FI PID eq {pid}" for pid, _, _ in pid_triplets)
```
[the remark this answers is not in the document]
antirotor: I've removed fallback options completely. Without psutil, the functionality was so limited anyway.

client/ayon_applications/process.py
antirotor: It is not a content, it is a path. And it is loaded in a separate way in the Manger tool.

iLLiCiTiT: Then the attribute name is not clear and confusing, should be `output_path`.

client/ayon_applications/ui/process_monitor.py
```diff
@@ -0,0 +1,1041 @@
...
+        # Delete output files
+        files_deleted = 0
+        for file_path in files_to_delete:
+            # File might not exist anymore, so we use contextlib.suppress
+            with contextlib.suppress(OSError):
+                os.remove(file_path)
+                files_deleted += 1
+        self.signals.finished.emit(deleted_count, files_deleted)
+
+    def _cleanup_single(self) -> None:
```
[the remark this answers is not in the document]
iLLiCiTiT: Why this was resolved?

client/ayon_applications/manager.py
```diff
@@ -524,26 +534,59 @@ def manager(self) -> ApplicationManager:
...
+            "name": self.application.full_name,
+            "cwd": os.getcwd(),
             "args": self.launch_args,
-            "env": app_env
+            "env": app_env,
+            "output": output_file,
```
iLLiCiTiT, line 586: Unused.
```suggestion
```

client/ayon_applications/manager.py
```diff
@@ -524,26 +534,59 @@ def manager(self) -> ApplicationManager:
...
+        ) as temp_file:
+            output_file = temp_file.name
+        # create temporary file to read back pid
+        with tempfile.NamedTemporaryFile(
+            mode="w", prefix="ayon_pid_", suffix=".txt", delete=False
+        ) as pid_temp_file:
+            pid_file = pid_temp_file.name
+
         json_data = {
+            "name": self.application.full_name,
```
iLLiCiTiT, line 582: Unused
```suggestion
```

PR #96, opened 2025-08-13
Chore: Remove launcher action logic

iLLiCiTiT:
## Changelog Description
Remove  'get_applications_action_classes' and action.py.

The logic is deprecated since ayon-core 1.4.0, package.py already requires that version so it should be safe to remove the logic.

## Testing notes:
Don't know how to test. All should be working ok. The functionality was used in launcher tool in ayon-core.


kalisp (approved):
Looks it is not used elsewhere anymore.

BigRoy (approved):
LGTM - couldn't find any other references.

PR #98, opened 2025-09-16
Actions: Actions on workfile entities

iLLiCiTiT:
## Changelog Description
Implemented dynamic actions for workfile entities.

Application actions are shown on workfile entities if they have filled host name. The applications still folow task types filtering.

Moved addon class from `__init__.py` to `addon.py`. Refactored app items preparation so it can be re-used at multiple places.

## Testing notes:
1. Application actions work the same way on tasks, follows the filtering set in settings (using task types).
2. Launcher tool can show application actions on workfiles (requires ayon-core https://github.com/ynput/ayon-core/pull/1434 ).


MilaKudr (approved):
image

PR #101, opened 2025-09-17
Feature: Launch app when task or folder is missing from context

Tilix4:
## Changelog Description
Allow to launch an application from a context without folder nor task.

This feature is based on [stax integration use case](https://community.ynput.io/t/stax-host-addon-architecture/2793): we want to open a folder which doesn't have a task, like an episode, in order to open blender in the episode context and load all shots from this episode.

## Testing notes:
1. Launch Blender using this code in console
```py
from ayon_core.addon import AddonsManager
manager = AddonsManager()
apps = manager.get_enabled_addon("applications")
apps.launch_application(
            "blender/4-2",
            project_name="your_project", # TODO
            folder_path="/episodes/se01/sh001", # TODO
        )
```


Tilix4:
I'll insist on this, blocking to basically launch an app on raw folders is a strong limitation. Avoiding workfile creation on folders I'd understand, but there are many use cases where we would like to open at a hierarchy folder context, and overviewing an editing is one of them.

MustafaJafar:
Adding my 2 cents, 
- I can imagine you want to have something similar in UX to tray publisher where you can launch without folder or task context and it should get the context from the context selection on the left panel. Also, you don't care about saving any workfiles.
- As far as I know, Launching apps without task can cause some issues, e.g. default work template expect a task name, publisher expect current task as the default task when creating publish instnances.

client/ayon_applications/utils.py
```diff
@@ -619,12 +619,18 @@ def prepare_context_environments(
...
-            "Host launch require folder and task context."
+        missing_context = []
+        if not folder_entity:
+            missing_context.append('folder')
+        if not task_entity:
+            missing_context.append('task')
+        log.warning(
+            f"Missing {', '.join(missing_context)} from context. "
+            "Skipping workfile preparation."
         )
```
iLLiCiTiT, line 632:
```suggestion
        log.warning(
            "Task is not set. Skipping workfile preparation."
        )
```

Tilix4:
I agree with most of @BigRoy statement and I'll contribute to the brainstorming.

> * Do not make it appear everywhere all the time, users will be doing it wrong

I'm not saying that every host should appear on every folder. Only that it should be possible to launch any app on a folder without a task in context, but **not by default in the launcher**. Maybe I wasn't explicit enough there, making it "possible" doesn't mean "accessible" all the time.

> * Perhaps an application should be able to define whether it's allowed to launch without an active task in context. So that it's special to those integrations only?

I think this isn't up to the application to decide. For example, Stax is a way of working with Blender, so if you prevent Blender to be launched from folders, you're blocking Stax to function.

My point is: All is about what kind of data you're about to create. Most of the time DCCs are used to create production data, which are meant to be used in the final render. In such case, tasks are the way to work.
If you want to create content about a data, metadata like, which is what review is about (it could be considered as a metadata about a file, about a version in Ayon's workflow), therefore you can open from folders and will create metadata for a bunch of files/representations/versions in a row. BUT this is not a matter of which DCC you're about to use, I agree for most DCCs this has no point and will never be considered though (imagine reviewing an editing in nuke...).

client/ayon_applications/utils.py
Tilix4: Shall I raise an error when a folder is not provided? I can imagine working without tasks but not without folders ATM.

iLLiCiTiT: Not sure about that. Considering what actually happens here in this function, it should be just:
```python
if not task_entity:
    log.warning(
        "Task is not set. Skipping workfile preparation."
    )
    return
```

iLLiCiTiT: You don't have to, the information has no value, only what matters is if is set task.

The message should be `"Task is not set. Skipping workfile preparation."` and it doesn't matter if folder is or is not set, the task is the important bit.

MustafaJafar (commented):
> This feature is based on [stax integration use case](https://community.ynput.io/t/stax-host-addon-architecture/2793): we want to open a folder which doesn't have a task, like an episode, in order to open blender in the episode context and load all shots from this episode.

> it should be possible to launch any app on a folder without a task in context

Hello,
Revisiting this PR. 
As far as I know, there's no such a concept as "**episode context**" in AYON.
A similar use case as yours is multi-shot workflow. one maysay: I want to open Hiero in "**episode context**" where I can publish all shots from an episode. 
Another use case is reviewing. one may say: I want to open openrv in the episode context and load all shots from this episode.

And for these use cases we can suggest: 
- Multi-shot workflow: to have an `edit` task below some folder where you can open hiero and publish.
- Review: you can create a list of product versions e.g. list of latest render product versions below your shots from an episode. and review that list.

So, I believe this PR is trying to introduce a new concept to AYON. I believe submitting a request on our [feedback portal](https://feedback.ayon.app/) for discussing this concept is a better place.

BigRoy:
Getting back to this - did the use case extend beyond Stax here @Tilix4 ? Because I feel like just having a dedicated launch method for Stax that doesn't go through the host integration at all may be the more trivial approach then?

Tilix4:
> having a dedicated launch method for Stax that doesn't go through the host integration at all may be the more trivial approach then?

Actually we chose to have a `review` task automatically created when launching stax on a collection of shots, like episode or sequence. We'll explore that path and see how it goes. Once we have something robust enough, we'll be glad to share it.

PR #102, opened 2025-09-18
Environemnts: Store app tools to env variable on launch

iLLiCiTiT:
## Changelog Description
Store used tools into `AYON_APP_TOOLS` environment variable.

The tool names are joined with a `;`. It is stored for information purposes to be stored to workfile entities (for example).

## Testing notes:
1. Launching of application will set up `AYON_APP_TOOLS` with used tools.

### DIFF
--- a/client/ayon_applications/utils.py
+++ b/client/ayon_applications/utils.py
@@ -453,12 +453,12 @@ def prepare_app_environments(
     # Add tools environments
     groups_by_name = {}
     tool_by_group_name = collections.defaultdict(dict)
-    used_tool_names = set()
+    used_tool_names = []
     for key in tools:
         tool = app.manager.tools.get(key)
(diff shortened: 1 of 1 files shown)


client/ayon_applications/utils.py
```diff
@@ -453,10 +453,12 @@ def prepare_app_environments(
     # Add tools environments
     groups_by_name = {}
     tool_by_group_name = collections.defaultdict(dict)
+    used_tool_names = set()
     for key in tools:
         tool = app.manager.tools.get(key)
         if not tool or not tool.is_valid_for_app(app):
             continue
+        used_tool_names.add(tool.full_name)
```
BigRoy, line 461: Are tools technically applied in an order? If so, would it be worth to keep this a list to be entirely accurate, or?

PR #103, opened 2025-09-25
Chore: cleanup workfiles startup env vars

kalisp:
## Changelog Description
Removed control of opening of Workfiles app with environment variable. Replaced by control in `ayon-core` `ayon+settings://core/tools/Workfiles/open_workfile_tool_on_startup` which provides more granular control.

## Testing notes:
1. all Adobe products and Harmony shouldn't have any `*WORKFILES_ON_LAUNCH` in Environments.

### DIFF
--- a/server/applications.json
+++ b/server/applications.json
@@ -876,7 +876,7 @@
         "harmony": {
             "enabled": true,
             "host_name": "harmony",
-            "environment": "",
+            "environment": "{}",
             "variants": [
                 {
                     "name": "22",


server/applications.json
```diff
@@ -876,7 +876,7 @@
         "harmony": {
             "enabled": true,
             "host_name": "harmony",
-            "environment": "{\n    \"AYON_HARMONY_WORKFILES_ON_LAUNCH\": \"1\"\n}",
+            "environment": "",
```
iLLiCiTiT, line 879:
```suggestion
            "environment": "{}",
```

kalisp: Server actually handles that, but agree that this is cleaner.

PR #108, opened 2025-10-14
Add option to disable output redirection

antirotor:
## Changelog Description
Adding option to disable output redirection of launched apps. This solves the issue with Terminal apps or can selectively enable output console for Houdini. This is set per variant and by default, redirection is enabled.

## Testing notes:
- Disable output redirection for terminal and run it
- Disable output redirection for Houdini on Windows (it should open the console)
- Check the status in Process Monitor. Output path should be empty

### DIFF
--- a/client/ayon_applications/process.py
+++ b/client/ayon_applications/process.py
@@ -171,6 +171,8 @@ class ProcessManager:
             process_info (ProcessInfo): Process handler to store.
 
         """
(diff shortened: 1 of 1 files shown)


MustafaJafar (approved):
I love this one!

image

image

Here's how the process monitor looks like if someone is curious.
image

iLLiCiTiT (approved):

PR #113, opened 2025-11-13
Workfile entities: Automatically fill host name

iLLiCiTiT:
## Changelog Description
Automatically fill host name in data based on workfile extension.

The auto-fill happens when is applications addon set to production, it also happens only once.

## Testing notes:
1. Upload addon to server.
2. Add addon to bundle.
3. Set the bundle to production.
4. Workfile entities should have filled host name (have icons in launcher tool).

### DIFF
--- a/server/addon.py
+++ b/server/addon.py
@@ -513,9 +513,8 @@ class ApplicationsAddon(BaseServerAddon):
     # --------------------------------------------
     async def _workfile_entities_auto_filled(self) -> bool:
         async for _ in Postgres.iterate(
-            """
-            SELECT * FROM public.addon_data WHERE addon_name = $1 AND key = $2 
-            """,
+            "SELECT * FROM public.addon_data"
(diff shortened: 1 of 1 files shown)


BigRoy:
> The auto-fill happens when is applications addon set to production, it also happens only once.

Out of curiosity, but aren't we able to do the reverse, if there is a workfile entity without "host_name" can't we do this assumption in the launcher/actions for those that miss the data and just have the extension fallback there?

That way the data on the server remain clears that they are lacking, etc.?
Like, what does the fact that it has `host_name` set really help? Is that used by the web actions?

It'd be way more transparent.

BigRoy:
Ah wait! The launcher UI doesn't show workfiles on disk - it only shows workfile entities. I thought this was supposed to also handle the host name case for just files that were found on disk in the context, but launcher does not show those at all.

BigRoy (approved):
Sounds fine to me - didn't test.

PR #124, opened 2026-01-15
add rez support to allow for full path exe evaluation

nebukadhezer:
after prelaunch hook execution

## Changelog Description
In order support https://github.com/AcademySoftwareFoundation/rez the applications addon requires a fullpath to be present before hook evaluation. As we can evaluate the rez env in a prelaunch hook we would make an exception for rez apps.
The identification is an env var AYON_REZ_PACKAGES, this will disable the find executable check before hook execution and find the executable after the hook execution in the modified env from the hook.

This is my first change to the applications addon, and this also requires a small change on core on the find-executable function to accept optional_paths in which to search.

This is a minimal working example, I know that rez support is not a high prio for ynput, still the changes in core and applications are minimal and would be a start.

## Testing notes:
1. use this branch in dev mode
2. use the same branch from core in devmode https://github.com/ynput/ayon-core/pull/1649
3. install and configure the addon https://github.com/dev-holobay/ayon-hbay-rez-manager to actually get rez and the hooks needed to make it work, more info is there on the readme

looking forward to your thoughts.

### DIFF
--- a/client/ayon_applications/hooks/pre_set_rez_env.py
+++ b/client/ayon_applications/hooks/pre_set_rez_env.py
@@ -1,0 +1,111 @@
+import os
+import subprocess
+import json
+
+from ayon_applications import PreLaunchHook, ApplicationLaunchFailed, \
+    LaunchTypes
+
+from ayon_core.lib.vendor_bin_utils import find_executable
+from ayon_applications.defs import ApplicationExecutable
+
+
+class PreLaunchSetRezEnv(PreLaunchHook):
+    """Add Rez packages to the launch environment.
+
+    It will merge the parent launch context environment with a rez resolved
+    environment defined by the `AYON_REZ_PACKAGES` environment variable.
+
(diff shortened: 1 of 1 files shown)


BigRoy:
> So finally added the launch hook here to... not sure why I had left it in the other repo, to make it a simple package this needs to be here... so if rez is on PATH this will work, and my rez installer is not needed.

Lovely - I think it serves as a great example like this (and nice to see my [rez gist](https://gist.github.com/BigRoy/4f34def92121b1782183c48fbf02abc0) actually worked fine. ;)

I do think that the majority of this shouldn't natively be in `ayon-applications`. It's so specific to rez (and a certain workflow) that I don't think it should be stacked in here.... but we should look into making sure `ayon-applications` can be used in a huge variety of manners - including supporting rez in a workflow like this.

So I'm all for tweaking what's needed to make other addons allow extending the behavior here. ;)

PR #127, opened 2026-01-20
tbd: allow json with comments and trailing commas

vincentullmann:
## Changelog Description
allow comments and trailing commas in environment configs 

code wise this is mostly just this (for now)
```py
try:
    import json5 as json
except ImportError:
    import json
```
but there are probably a few more places that might need minor adjustments.

I'm just putting this PR out here as a wip / idea up for discussion

image

### DIFF
--- a/server/settings.py
+++ b/server/settings.py
@@ -1,5 +1,4 @@
 import os
-import json
 from pydantic import validator
 
 from ayon_server.addons import BaseServerAddon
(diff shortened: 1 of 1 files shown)


BigRoy:
From a usability standpoint, I can see this make sense.

I am rather concerned about the point made on the official `json5` package though - that it is _slow_.

> It has been reasonably well-tested, but it is SLOW. It can be 1000-6000x slower than the C-optimized JSON module, and is 200x slower (or more) than the pure Python JSON module.

The usability only affects the admin, the speed difference affects all places where we load these settings and apply them, which is widespread. With that, I'm not sure the gains are worth the cons?

Overall we're not using massive JSONs usually, yet still - in AYON Applications we're dealing with quite a few of the JSON fields, as such - I'm not convinced we always want to unpack them.

Also the fact that we're testing for `json5` on the server as a dependency does not mean the client side will have it - and I do still think it's the client side who does the decoding, right? (although I haven't checked it now)

PR #128, opened 2026-01-28
Add "Open source workfile" loader action

### LINKED ISSUE #137: YN-0441: Open Source Workfile action in the Loader
### Please describe the feature you have in mind and explain what the current shortcomings are?

Use case:

When reviewing renders in the Loader, artists want to quickly open the original workfile (Nuke/Maya/Silhouette scene) that created that render.

  

Technical approach:

Versions already store the source path in attrib.source (e.g., "{root[work]}/project/sq001/Shot01/work/paint/file_v001.nk")

  

Create a loader action plugin that resolves the path via anatomy and launches with the appropriate DCC

Similar to existing workfile launcher logic. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1465578744960909335/1465578750921019510)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0441]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=f95bed50fb4211f085ae8d5e6239e938)

### LINKED ISSUE #1635: YN-0365: Go to workfile from published content
### Please describe the feature you have in mind and explain what the current shortcomings are?

**Question:**

  

Is it possible to load published workfiles from the Loader and/or Launcher? Similar to the option in the "Work Files" Dialog but as part of the other widgets

  

**to add some context:** this request initially came up when looking at some asset copied using the "Push to Project" Action. The Action does copy over the published workfiles, which is great, but we've noticed that there is no way to load them from the launcher/loader. I then asked myself if this should be a general request of accessing published workfiles (not only for pushed products, but any published element)

  

**Possible implementation:**

  

For every publish that has a source context set that is essentially a workfile, we could have a "Show source in Workfiles" type of action which would essentially open the workfiles tool with that workfile selected. The idea behind that is that it'd work on any publish that came from a workfile, so that e.g. if you have a pointcache v005 somewhere, that you can easily find the source workfile that created it.

This way, if you wanted to quickly open the workfile to investigate it, you can.

For published workfiles belonging to a workfile it could even offer opening the "Source workfile" or "Copy published workfile and open" to pick either the source or published workfile to be selected in the publisher UI.

(It's slightly more complicated in the fact that publishes may not have a workfile, e.g. from tray publisher, or command-line, etc.) 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1458812577726468179)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0365]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=3b6a69b0ed3311f0a03683387f770da7)

BigRoy:
## Changelog Description

Add action to open source workfile that created the publish, directly from the standalone browser

image

image

_is it me or should the cancel and open buttons be swapped around?_

This may solve ticket [YN-0365](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=3b6a69b0ed3311f0a03683387f770da7) and could act as a usable fix for https://github.com/ynput/ayon-core/issues/1635 even though it's not exactly the same.

Quite a few TODOs in the code to discuss stilll, hence **draft** for now - yet still good to discuss overall approach.

Fix #137 

Fix: https://github.com/ynput/ayon-core/issues/1635

## Testing notes:

1. Open Tray browser
2. Try to open source workfile for any existing publishes.


client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,283 @@
...
+        # Only allow if no registered host, like in standalone browser
+        if self.host_name:
+            return False
+
+        if not selection.versions_selected():
+            return False
+
+        for version in selection.get_selected_version_entities():
+            if version.get("attrib", {}).get("source"):
+                return True
```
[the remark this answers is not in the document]
BigRoy: Let's try to live with:

- `if not source: return False`
- `if source.startswith("{root"): ...`
- `elif os.path.exists(source): ...` then assuming it's a valid source workfile?

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,284 @@
...
+        for i, app in enumerate(apps):
+            label = app.full_label or app.name
+
+            # Highlight the app that was used to create the publish so that
+            # the user knows it's the recommended one to open with.
+            if source_app_full_name and app.full_name == source_app_full_name:
+                preferred_index = i
+                label += " (used to create publish)"
+
+            # TODO: Show application icon if available
```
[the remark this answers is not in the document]
moonyuet: It might be reused for this case as it is already showing the list of applications we can. @iLLiCiTiT what do you think? Or we're better sticking with ActionForm instead?

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,283 @@
...
+    def _launch_app(
+        self,
+        app: "Application",
+        version: dict,
+        workfile_path: str,
+        project_name: str,
+        addons_manager: AddonsManager,
+    ):
+        """Launch application with workfile."""
+
```
iLLiCiTiT:
```suggestion
    run_ayon_launcher_process(
        "addon", "applications", "launch",
        "--project", project_name,
        "--folder", folder_path,
        "--task", task_name,
        "--app", app.full_name,
        "--use-last-workfile", "0",
    )
```

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,283 @@
...
+    LoaderActionResult,
+)
+from ayon_core.style import load_stylesheet
+from qtpy import QtWidgets, QtCore
+
+if TYPE_CHECKING:
+    from ayon_applications.manager import Application
+
+
+class OpenSourceWorkfileAction(LoaderSimpleActionPlugin):
```
iLLiCiTiT, line 30: BTW this would not work with project bundle. It would have to be webaction.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
moonyuet: This is not working to launch the source workfile. 
image
image

BigRoy: Yup - it's lacking a `--workfile-path` argument to the `launch` CLI entry point. Once that's there - this should perhaps work?

I think we should add an optional `--workfile-path` argument that overrides `--use-last-workfile` and when passed is used as the exact workfile to open.

@iLLiCiTiT would that technically already solve the project bundle aspect since it'd go through the launcher, etc.?

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,283 @@
...
+                exe = app.find_executable()
+            except Exception:
+                continue
+
+            if exe and os.path.exists(str(exe)):
+                compatible.append(app)
+
+        return compatible
+
+    def _show_app_dialog(
```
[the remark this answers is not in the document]
BigRoy: If that can work - sounds good to me.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
BigRoy: The action should be able to trigger an app launch into the valid project bundle. Is the only reason it's currently not working just because we're not going through CLI, or?

As in, would this fix it: https://github.com/ynput/ayon-applications/pull/128/changes#r3264305267

client/ayon_applications/plugins/load_actions/open_source_workfile.py
iLLiCiTiT: > would that technically already solve the project bundle aspect since it'd go through the launcher, etc.?

Partially. The list of available applications can be different in project bundle, so you can send in `maya/2027` but there is no `maya/2027` in the project bundle settings.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
iLLiCiTiT: The list of applications can be different. If you're using `ApplicationsManager` to get list of applications it is not 100% project bundle support.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
iLLiCiTiT: Regarding the argument, do we want to allow open the "real" published file? Created PR for that https://github.com/ynput/ayon-applications/pull/155 but we should consider if it is ok.

BigRoy: > Regarding the argument, do we want to allow open the "real" published file?

No the focus should not be at doing that - perhaps in the case of loading published workfile it'd require the copy & open. But this action _should_ be opening the workfile source though? Or?

client/ayon_applications/plugins/load_actions/open_source_workfile.py
BigRoy: > The list of applications can be different. If you're using ApplicationsManager to get list of applications it is not 100% project bundle support.

I see - that makes sense. Can we make it display the right ones instead? 🤔

iLLiCiTiT: Right now there is https://github.com/ynput/ayon-applications/pull/139 I can add one endpoint for the project so it will get all available applications, but that won't have information about extensions...

BigRoy: > Right now there is https://github.com/ynput/ayon-applications/pull/139 I can add one endpoint for the project so it will get all available applications, but that won't have information about extensions...

That's because it may not access that version of the addon in the current bundle? 🤔 

Perhaps we shouldn't overcomplicate things. The _best_ case scenario is that it's able to at least find the relevant source application in the versions metadata - and with that by default pick the _actual_ valid application to use from settings as the pre-selected entry. I'd rather have it show an app version too many, than none showing at all so maybe just taking from studio settings is the valid approach here? (Was applications addon even overridable per-project?)

client/ayon_applications/plugins/load_actions/open_source_workfile.py
BigRoy: #155 is merged - so this should now be able to take `--workfile-path` argument.

moonyuet: Already implemented the change in the latest commit.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
iLLiCiTiT: > Perhaps we shouldn't overcomplicate things. 

If we do consider project bundles, we have to do more...

> and with that by default pick the actual valid application to use from settings as the pre-selected entry.

That's the point related to project bundles, you can't do it client side because you don't know which settings to use, and you don't know the structure of settings of other applications addon versions. That's why https://github.com/ynput/ayon-applications/pull/139 was added, so you just get information about available applications no matter which version of applications addon is used.

>  I'd rather have it show an app version too many, than none showing at all so maybe just taking from studio settings is the valid approach here?

Won't work, project bundles do work as if you'd freeze studio settings too. That would work with first implementation of project bundles, not with current implementation.

> Was applications addon even overridable per-project?

Right now all "our cases" should work, which means this would break the assumption (maybe except some private addons that are not that maintained). That's why I added https://github.com/ynput/ayon-applications/pull/139 so we can actually do something about it (e.g. ftrack is getting webactions for the selection to filter out applications actions ->that is "not ideal").

moonyuet: @iLLiCiTiT Just want to clarify a little bit. I saw the #139 and you mentioned https://github.com/ynput/ayon-applications/pull/151 is better solution. Does it mean we can use the getter functions from there instead of #139?

iLLiCiTiT: > Does it mean we can use the getter functions from there instead of https://github.com/ynput/ayon-applications/pull/139?

You can't use the methods from https://github.com/ynput/ayon-applications/pull/151 based on bundle you're currently in, you have to resolve it for the bundle in which the application will be launched.

So you either have to resolve which addon version has to be used on your own and then use the mothods on the addon version (and the methods are not available on older versions), or use that PR.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,219 @@
...
+
+        allowed_names = {
+            item["full_name"]
+            for item in app_items
+            if item.get("host_name") in host_names
+        }
+        if not allowed_names:
+            return []
+
+        # 3) map back to Application objects for choose_app and launch flow
```
iLLiCiTiT: The point is that you don't use applications addon from current session to launch the application. You have to do something like this.
```python
run_detached_ayon_launcher_process(
    "addon", "applications", "launch",
    "--app", full_app_name,
    "--project", project_name,
    "--folder", folder_path,
    "--task", task_name,
)
```

Now, I don't know how to get the project, folder and task, but you can create fake `Application` objects for `choose_app` dialog like this.
```python
def _create_fake_applications(app_items: list[dict[str, Any]]) -> list[Application]:
    app_items_by_group = collections.defaultdict(list)
    for app_item in app_items:
        full_name = app_item["full_name"]
        group_name = full_name.split("/")[0]
        app_items_by_group[group_name].append(app_item)

    output = []
    for group_name, group_app_items in app_items_by_group.items():
        host_name = None
        variants = []
        for app_item in group_app_items:
            variants.append({
                "name": app_item["full_name"].split("/")[1],
                "label": app_item["variant_label"],
                "environment": "{}",
                "arguments": [],
                "executables": {},
            })
        group = ApplicationGroup(
            group_name,
            {
                "enabled": True,
                "environment": "{}",
                "host_name": host_name,
                "variants": variants,
            },
        )
        for variant in group.variants.values():
            output.append(variant)
    return output
```

Or we can modify the dialog to allow to pass dataclass with less information than what Application needs to be created.

Point is that `apps_addon` should not be needed at all. You also can't validate the executable. For the executable validation we'd have to add that information to the app items too.

Or we can add dedicated cli for that part of logic.

moonyuet: I actually prefer adding dedicated cli for long term and it might be reused in some loader actions if they are required to load the application addons. 
@BigRoy thoughts?

BigRoy: For what's the dedicated CLI entry point needed?

- We already have a CLI to launch apps, that's the one e.g. @iLLiCiTiT pointed out at [the top here](https://github.com/ynput/ayon-applications/pull/128#discussion_r3520136076).
- Then all we need to do is allow the user to pick the right app - and then launch through CLI, right @iLLiCiTiT ?

client/ayon_applications/plugins/load_actions/open_source_workfile.py
BigRoy: Preferably use that PR then I suppose :')

client/ayon_applications/plugins/load_actions/open_source_workfile.py
iLLiCiTiT: > Then all we need to do is allow the user to pick the right app

Probably yes, if you don't have available executables it will show you a message on launch.

iLLiCiTiT: What is missing is the filtering of applications by extension. You really can't do that because currently available addons are NOT those this might be used in the target project. So we would have to show all applications in the apps selection. The whole step 1 and 3 can't happen in this process.

BigRoy: > What is missing is the filtering of applications by extension. You really can't do that because currently available addons are NOT those this might be used in the target project. So we would have to show all applications in the apps selection.

I see. Regardless, let's filter them for now by the current project's addons (with the side effect that project bundles may be differing) and then using the apps defined of the target project. (For as far as that's possible?) YES, there _may_ be a mismatch, but in practice these extensions should be relatively static across updates and probably apps and addons even though they might differ in version there's a relatively high chance a lot of the same addons are enabled, etc.

However, we'll need to add a checkbox in the app picker UI to disable the extensions filtering and instead show ALL apps (by default it's filtered to only show those matching by extension.) so that for those edge cases, the user can disable and still pick from the full list.

We'll also need a search bar to search among the large list of app entries, otherwise it'll be very uncomfortable.

iLLiCiTiT: > YES, there may be a mismatch, but in practice these extensions should be relatively static across updates and probably apps and addons even though they might differ in version there's a relatively high chance a lot of the same addons are enabled, etc.

But you can have enabled different addons.

BigRoy: > But you can have enabled different addons.

You can, yes - but we'll make the assumption that the likelihood of that is _relatively_ small so that the best UX is still to show filtered what we can - but allow the user to display all in the scenarios that it does differ.

iLLiCiTiT: > You can, yes - but we'll make the assumption that the likelihood of that is relatively small so that the best UX is still to show filtered what we can - but allow the user to display all in the scenarios that it does differ.

BTW It is not just about extensions. You can't validate the executables, because you might get applications you don't have available in your current settings, or it is not filled in your current settings. Everything that can be different in project bundle (which is everything settings related) can't be handled in current process, unless current process is already in the context of the project we're looking at -> which is not if you're in tray browser, or if you look into library project.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,219 @@
...
+        if not file_ext:
+            return LoaderActionResult(
+                f"Version source '{workfile_name}' has no extension.",
+                success=False,
+            )
+
+        # Get compatible applications
+        task_id = version.get("taskId")
+        project_name = selection.project_name
+        addons_manager = AddonsManager()
```
iLLiCiTiT:
```suggestion
        addons_manager = self.get_addons_manager()
```

client/ayon_applications/plugins/load_actions/open_source_workfile.py
BigRoy: I see, so we should really preferably skip the `find_executable` check in the choose dialog? 🤔

moonyuet: > I see, so we should really preferably skip the `find_executable` check in the choose dialog? 🤔

I removed the check on [36937da](https://github.com/ynput/ayon-applications/pull/128/commits/36937daa0a0d7e2b276381d067cf028af515f227) and improve [bd9aca8](https://github.com/ynput/ayon-applications/pull/128/commits/bd9aca8af9f43530b3e374835768361a30324258)

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,219 @@
...
+
+        # Find the applications matching the host names
+        apps_addon = addons_manager.get("applications")
+        if not apps_addon:
+            return []
+
+        app_items = apps_addon.get_application_items(
+            project_name,
+            task_id=task_id,
+            version=apps_addon.version
```
iLLiCiTiT: You want to get applications for the project and task, not for current applications addon.
```suggestion
```

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,219 @@
...
+            if file_ext in extensions:
+                host_names.add(host_name)
+
+        if not host_names:
+            return []
+
+        # Find the applications matching the host names
+        apps_addon = addons_manager.get("applications")
+        if not apps_addon:
+            return []
```
iLLiCiTiT: This should be moved before the collection of host names. Or removed, if applications addon is not available then you won't be able to discover this action.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
iLLiCiTiT: But you're still using applications from current applications addon.

You'll get applications that are available for the project and task, the project might have different applications set in it's settings and they won't be available in current applications addon. That's why I wrote the first comment of this conversation https://github.com/ynput/ayon-applications/pull/128/changes#r3520136076 .

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,218 @@
...
+    def _get_compatible_apps(
+        self,
+        addons_manager,
+        file_ext,
+        project_name,
+        task_id,
+    ) -> list[Any]:
+        """Get compatible applications for file extension."""
+
+        # 1) host names that can open this extension
```
iLLiCiTiT:
```suggestion
        # 1) host names that can open this extension
        # NOTE: Does not respect project bundle addons.
```

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,218 @@
...
+                success=False,
+            )
+        # Launch application
+        try:
+            product_id = version["productId"]
+            product = ayon_api.get_product_by_id(project_name, product_id)
+            folder = ayon_api.get_folder_by_id(
+                project_name, product["folderId"]
+            )
+            task = (
```
iLLiCiTiT: We don't allow to launch application without task. And using `None` in the arguments will cause crash.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
moonyuet: Are we gonna do the dedicated cli for that or we can use the fake application object with detach run on ayon launcher to get the correct applications?

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,218 @@
...
+        """Check if any selected version has source workfile."""
+        # Only allow if no registered host, like in standalone browser
+        if self.host_name:
+            return False
+
+        if not selection.versions_selected():
+            return False
+
+        for version in selection.get_selected_version_entities():
+            source = version.get("attrib", {}).get("source")
```
iLLiCiTiT, line 59:
```suggestion
            if not version["taskId"]:
                continue

            source = version.get("attrib", {}).get("source")
```

client/ayon_applications/plugins/load_actions/open_source_workfile.py
moonyuet: [4b91332](https://github.com/ynput/ayon-applications/pull/128/commits/4b913325a4ec43012064bd90759fb4314bcc287b)

client/ayon_applications/plugins/load_actions/open_source_workfile.py
moonyuet: [0a0c34f](https://github.com/ynput/ayon-applications/pull/128/commits/0a0c34f8a4947280422636aedb4d0de606959b67)

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,218 @@
...
+            )
+            run_ayon_launcher_process(
+                "addon", "applications", "launch",
+                "--project", project_name,
+                "--folder", folder["path"],
+                "--task", task["name"] if task else None,
+                "--app", selected_app.full_name,
+                "--workfile-path", workfile_path,
+                "--use-last-workfile", "0",
+            )
```
iLLiCiTiT:
```suggestion
            run_ayon_launcher_process(
                "addon", "applications", "launch-by-id",
                "--project", project_name,
                "--task-id", version["taskId"],
                "--app", selected_app.full_name,
                "--workfile-path", workfile_path,
                "--use-last-workfile", "0",
            )
```

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,218 @@
...
+        workfile_name = os.path.basename(source_path)
+        file_ext = os.path.splitext(workfile_name)[1].lower()
+        if not file_ext:
+            return LoaderActionResult(
+                f"Version source '{workfile_name}' has no extension.",
+                success=False,
+            )
+
+        # Get compatible applications
+        task_id = version.get("taskId")
```
iLLiCiTiT:
```suggestion
        task_id = version["taskId"]
```

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,211 @@
...
+            None
+        )
+
+        if not selected_app:
+            return LoaderActionResult(
+                f"Selected application '{selected_app_name}' was not found.",
+                success=False,
+            )
+        # Launch application
+        try:
```
iLLiCiTiT: The try except block should not be needed.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
iLLiCiTiT: I guess faking it will allow us to use it even with older applications addon.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
moonyuet: [43ac865](https://github.com/ynput/ayon-applications/pull/128/commits/43ac865d06f21be44b455083a5c21cc885aa97b8)

client/ayon_applications/plugins/load_actions/open_source_workfile.py
moonyuet: > I guess faking it will allow us to use it even with older applications addon.

I implement the fake application object, and run detach ayon launcher process with [4683b4f](https://github.com/ynput/ayon-applications/pull/128/commits/4683b4fac2dbea8340cf3ebba605b6e3b0b5451b).

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -108,9 +108,10 @@ def execute_simple_action(
                 success=False,
             )
         apps_addon = addons_manager["applications"]
+        ayon_app_name = version["data"].get("ayon_app_name")
         selected_app = next(
             (app for app in compatible_apps
-             if app.full_name == version["data"]["ayon_app_name"]),
+             if ayon_app_name and app.full_name == ayon_app_name),
             None
         )
```
BigRoy: This could now be in a `if ayon_app_name` indentation, because if it's not set we can iterate all compatible apps but it would never return anything. So doesn't make sense to iterate then.

moonyuet: > This could now be in a `if ayon_app_name` indentation, because if it's not set we can iterate all compatible apps but it would never return anything. So doesn't make sense to iterate then.

Ignore my comment just now. I will take a look at it tmr.

moonyuet: [157c164](https://github.com/ynput/ayon-applications/pull/128/commits/157c164d9200308b17db8af458b7de5a03f9d45a) I have done some cleanup to increase readability.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,248 @@
...
+            return LoaderActionResult(
+                f"No compatible applications found for {file_ext}",
+                success=False,
+            )
+        app_by_name = {app.full_name: app for app in compatible_apps}
+        selected_app = None
+        ayon_app_name = version["data"].get("ayon_app_name")
+        if ayon_app_name:
+            selected_app = app_by_name.get(ayon_app_name)
+        if not selected_app:
```
BigRoy, line 133: I have my doubts here - it makes it very quick to use to quickly open it with the most likely correct app - but it also gives you NO way to actually try and open it with a different compatible application. 🤔

Which makes me think whether we should instead always show the prompt to the user, but have the matching version pre-selected if it has it. 🤔 

Or we allow via Option Box to enable "Choose application manually" so the user always get the prompt, regardless of whether it found a quick match.

The latter might be the nicest because it's very likely that a user will indeed want to open it with the original application.

client/ayon_applications/plugins/load_actions/open_source_workfile.py
```diff
@@ -0,0 +1,248 @@
...
+            return []
+        # host names that can open this extension
+        # NOTE: Does not respect project bundle addons.
+        host_names: set[str] = set()
+        for addon in addons_manager.addons:
+            if not isinstance(addon, IHostAddon):
+                continue
+
+            try:
+
```
iLLiCiTiT, line 192:
```suggestion
```

client/ayon_applications/plugins/load_actions/open_source_workfile.py
moonyuet: >Or we allow via Option Box to enable "Choose application manually" so the user always get the prompt, regardless of whether it found a quick match.

But we need to be careful on this. Some DCC cannot open workfile which is saved in newer version of the applications. e.g. 3dsMax.

moonyuet: [84d80f2](https://github.com/ynput/ayon-applications/pull/128/commits/84d80f2273bb3a78793934658997735237dc665e)

BigRoy: > But we need to be careful on this. Some DCC cannot open workfile which is saved in newer version of the applications. e.g. 3dsMax.

When you start picking your own DCC choice - then that's up to your own responsibility I'd say.

PR #139, opened 2026-04-13
Chore: Applications action helpers

iLLiCiTiT:
## Changelog Description
Added helper functions to get list of available application actions for a given context.

These functions can be used to receive possible actions with all necessary information for a certain context. This will be used in ftrack addon. Also affects existing function `get_applications_for_context` which returns list of available application names.

The new api functions can be used by other plugins. The implementation does handle project bundles on it's own and auto-resolves the settings for addons that don't have the api methods implemented yet.

## Testing notes:
1. Can be tested with project bundles where different actions can be defined from studio settings.
2. Use the functions to get available applications for the project.


moonyuet (approved):
It's all working now. See the screenshot.

Image

PR #143, opened 2026-04-22
Group label per variant

iLLiCiTiT:
## Changelog Description
Added option to change group label per variant.

This gives option to split Houdini variants into their own categories as needed.

## Testing notes:
1. Define custom group label on an application variant.
2. The action in UI should show the group (e.g in launcher).

Resolves https://github.com/ynput/ayon-applications/issues/142


MilaKudr (approved):
It works, but the hint/tooltip is a little confusing to me.

Image

PR #146, opened 2026-04-30
Actions: Allow to ungroup application actions

iLLiCiTiT:
## Changelog Description
Application actions can be ungrouped in UI.

Added `show_grouped` in settings which can be used to disable grouping of the actions. If is disabled then the action will not be grouped, that is achieved by not defining group label but only label.

### Screenshot
image 
image

## Testing notes:
1. Create package, upload to server and use in bundle.
2. Choose application with multiple variants and uncheck `Group in UI` for one of the variants.
3. Open launcher UI.
4. The action of the variant should not be showed as a group.

### DIFF
--- a/server/actions.py
+++ b/server/actions.py
@@ -73,6 +73,7 @@ def get_items_for_app_groups(groups):
                 "group_label": variant_group_label,
                 "variant_label": variant_label,
                 "icon": icon,
+                "show_grouped": variant["show_grouped"],
             })
 
     items.sort(key=_sort_getter)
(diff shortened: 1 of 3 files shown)


BigRoy (approved):
It works, the settings are rather versbose - I'd prefer the checkbox if somehow possible to be left of the Group Label (somehow joined together) or have Group label visually be hidden if Grouping is disabled, but it's cosmetics really.

Functionally this works as advertised. 👍

iLLiCiTiT:
> It works, the settings are rather versbose - I'd prefer the checkbox if somehow possible to be left of the Group Label (somehow joined together) or have Group label visually be hidden if Grouping is disabled, but it's cosmetics really.

I've explained that already, but the group label is NOT directly connected. You might want change group label from `Houdini` to `Houdini FX` and disable grouping in UI at the same time (change group label and disable grouping). That also gives you the option to swap variant and group labels `Houdini 21` -> `21 Houdini` so you see version first.

PR #147, opened 2026-05-05
Settings: Name validations

iLLiCiTiT:
## Changelog Description
Added simple validation of name values of applications and tools names.

Make sure that applciations and tools names are not empty and do not contain `/` which is used for full name.

### Possible other issues/validations
- I could image `&` character could be also an issue for cli.
- Not sure if there are regex restrictions for webaction topic value (the name is part of the topic).

## Testing notes:
Validate code changes.
1. Using empty name or name with `/` should not allow to save settings.

### DIFF
--- a/server/settings.py
+++ b/server/settings.py
@@ -226,6 +226,17 @@ class AppVariant(BaseSettingsModel):
         )
     )
 
+    @validator("name")
+    def validate_name(cls, value):
+        if not value:
+            raise BadRequestException("Application variant is empty")
+
+        if "/" in value:
+            raise BadRequestException(
+                f"Application variant ({value}) can't contain '/'"
+            )
+        return value
+
     @validator("environment")
     def validate_json(cls, value):
         return validate_json_dict(value)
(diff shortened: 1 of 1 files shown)


server/settings.py
```diff
@@ -226,6 +226,17 @@ class AppVariant(BaseSettingsModel):
...
     )
 
+    @validator("name")
+    def validate_name(cls, value):
+        if not value:
+            raise BadRequestException("Application variant is empty")
+
+        if "/" in value:
+            raise BadRequestException(
+                f"Application variant ({value}) can't contain '/'"
```
BigRoy: Why the ()?
```suggestion
                f"Application variant '{value}' can't contain '/'"
```

BigRoy (approved):
Code seems fine, but beware... if someone happened to have settings saved previously with e.g. an empty tool name (they just pressed the + but didn't populate it) then the settings will now fail to load. Essentially meaning that the user will have lost its existing data.

This happened before when we added e.g. the validator to the environment JSON fields that it's actually valid JSON - when someone had saved settings that were invalid, it'd essentially break their settings after the validation got added.

So this _may_ be painful 🗡️

server/settings.py
iLLiCiTiT: The error is `"Application variant can't contain '/'"`, the value is helper information.

PR #151, opened 2026-05-12
Server addon api: Methods to get available apps and tools

iLLiCiTiT:
## Changelog Description
Proposal to be able to get apps and tools available for the addon versoin.

Added 5 new public methods to the addon. Methods `get_application_items` and `get_tool_items` to get the items for the version of the addon. Helper method `get_addon_for_context` to get application addon version object for passed context based on project and variant. And methods `get_applications_for_context` and `get_tools_for_context` which will find addon version based on the context and call `get_application_items` or `get_tool_items` on them -> that will start to work from this version onwards.

It is not really possible to support older versions as there is no way to access their code if is not part of addon classes. Without this we won't be able to do that even in future, so sooner this is merged the earlier we're able to use it.

The idea is that this can be used by other addons e.g. in settings or action forms.

## Testing notes:
Validate code changes.

It should be possible to get available tools and applications from other addons.

### DIFF
--- a/server/addon.py
+++ b/server/addon.py
@@ -62,7 +62,11 @@ if TYPE_CHECKING:
     )
 
 from .constants import LABELS_BY_GROUP_NAME
-from .settings import ApplicationsAddonSettings, DEFAULT_VALUES
+from .settings import (
+    ApplicationsAddonSettings,
+    DEFAULT_VALUES,
+    applications_enum,
+)
 from .utils import (
     ApplicationItem,
     ToolItem,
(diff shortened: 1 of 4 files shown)


server/addon.py
```diff
@@ -342,6 +356,141 @@ async def convert_settings_overrides(
...
+    ):
+        """Helper that can be used to get applications enum for settings.
+
+        Example:
+            from ayon_server.addons import AddonLibrary
+
+            async def apps_enum(project_name, addon, settings_variant):
+                addon_library = AddonLibrary.getinstance()
+                app_addons = addon_library.data.get("applications") or {}
+                for addon in app_addons.values():
```
BigRoy: `.values()` does not exist on `data`
```suggestion
                for addon in app_addons.versions.values():
```

BigRoy (approved):
The enum seems to work. Tested in: https://github.com/ynput/ayon-usd/pull/124

The example however in the docstring for the apps enum resolver however is incorrect.

Also, I ended up using `app_addons.latest` instead of iterating the versions so that it'd always call the method from the latest app addons version.

server/addon.py
```diff
@@ -342,6 +356,141 @@ async def convert_settings_overrides(
...
+                    if not hasattr(addon, "get_applications_settings_enum"):
+                        continue
+                    return await addon.get_applications_settings_enum(
+                        project_name=project_name,
+                        settings_variant=settings_variant,
+                    )
```
iLLiCiTiT:
```suggestion
                addon = app_addons.latest
                if hasattr(addon, "get_applications_settings_enum"):
                    return await addon.get_applications_settings_enum(
                        project_name=project_name,
                        settings_variant=settings_variant,
                    )
```

PR #154, opened 2026-05-19
Launch cli: Allow to pass any workfile path to launch

iLLiCiTiT:
## Changelog Description
It is possible to pass workfile path to launch commands.

## Additional review information
This might be dangerous as it allows to open any workfile out of workdir.

## Testing notes:
1. Passing workfile path does launch the workfile.

PR #155, opened 2026-05-19
Launch cli: Allow to pass any workfile path to launch

iLLiCiTiT:
## Changelog Description
It is possible to pass workfile path to launch commands.

This might be dangerous as it allows to open any workfile out of workdir. PR was created because of https://github.com/ynput/ayon-applications/pull/128 . We have to validate if we want to allow this.

Re-opened https://github.com/ynput/ayon-applications/pull/154 .

## Testing notes:
1. Passing workfile path does launch the workfile.


BigRoy (approved):
Yes, we should allow the CLI to open with any file - regardless of where it exists. If someone is to open published workfiles - whip them... but I don't see why the API should disallow picking a dedicated workfile to open.

